PR #22, opened 2014-06-02
[WIP] Macro "extends" auto mode: removed dependency on the presenter

matej21:
What do you think?

- [ ] Fix coding style, add copyright etc.
- [ ] Fix failing tests
- [ ] Create ILayoutFileFinder implementation in nette/application

### DIFF
--- a/src/Latte/ILayoutFileFinder.php
+++ b/src/Latte/ILayoutFileFinder.php
@@ -1,0 +1,16 @@
+<?php
+namespace Latte;
+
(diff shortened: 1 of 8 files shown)


f3l1x:
I like an idea. 

@ping @dg @matej21 @hrach 

How do you think .."This should be solved via Loader..."?

dg:
It would be useful to find a nice way how to pass these special variables, other way than underscored names.

PR #29, opened 2014-08-16
[WIP] snippets refactoring

matej21:
Snippets are now completely independent on the nette/application. I will send related PR to the nette/application in few minutes/hours.

What do you think?


dg:
@matej21 do you think it can be incorporated in 2.4?

matej21:
@dg I have some ideas how to rewrite it a bit. But I'm not sure when I find the time :\ (maybe this weekend)

matej21:
@dg reworked. One test is failing because of bad indent (not sure which indent is right, probably "actual" :) will fix later). nette/application#22 is also updated.

dg:
Wow!

matej21:
the indent is fixed. but there is a small BC break - previous behavior was a bit inconsistent and whitespaces on the last line were removed only for a snippet macro.

f3l1x:
Good job @matej21! :+1: It looks more readable.

src/Latte/ISnippetBridge.php
```diff
@@ -0,0 +1,57 @@
...
+	 * @return bool
+	 */
+	public function isSnippetMode();
+
+
+	/**
+	 * @param string
+	 * @return bool
+	 */
+	public function isInvalid($name);
```
dg: `needsRedraw`?

src/Latte/SnippetDriver.php
```diff
@@ -0,0 +1,102 @@
+<?php
```
dg: This should be IMHO in Latte\Runtime

src/Latte/Macros/BlockMacros.php
```diff
@@ -201,17 +201,18 @@ public function macroBlock(MacroNode $node, PhpWriter $writer)
 				}
 				$parent->data->dynamic = TRUE;
 				$node->data->leave = TRUE;
-				$node->closingCode = "<?php \$this->global->dynSnippets[\$this->global->dynSnippetId] = ob_get_flush(); ?>";
+				$node->closingCode = "<?php \$this->global->snippetDriver->leave(); ?>";
+				$enterCode = "\$this->global->snippetDriver->enter(" . $writer->formatWord($name) . ", \\Latte\\SnippetDriver::TYPE_DYNAMIC);";
```
dg: here are single quotes better

src/Latte/Macros/BlockMacros.php
```diff
@@ -294,18 +295,14 @@ public function macroBlockEnd(MacroNode $node, PhpWriter $writer)
 				$node->content = $node->innerContent;
 			}
 
+			if (($node->name === 'snippet' || $node->name === 'snippetArea') && strpos($node->data->name, '$') === FALSE) {
+				$type = $node->name === 'snippet' ? 'TYPE_STATIC' : 'TYPE_AREA';
+				$node->content = '<?php $this->global->snippetDriver->enter('
```
dg: what about `$type = $node->name === 'snippet' ? Latte\SnippetDriver::TYPE_STATIC : Latte\SnippetDriver::TYPE_AREA;`

src/Latte/SnippetDriver.php
```diff
@@ -0,0 +1,102 @@
...
+class SnippetDriver
+{
+	use Strict;
+
+	const TYPE_STATIC = 'static',
+		TYPE_DYNAMIC = 'dynamic',
+		TYPE_AREA = 'area';
+
+	/** @var array */
+	protected $stack = [];
```
dg: better is to make everything private unless somebody need to make it protected

src/Latte/Template.php
```diff
@@ -173,13 +173,18 @@ protected function initialize(& $params)
 				$this->renderBlock($this->params['_renderblock'], $this->params);
 				return TRUE;
 			}
+		} elseif (isset($this->global->snippetBridge) && !isset($this->global->snippetDriver)) {
```
dg: Do you think that support for `$this->global->snippetBridge` is necessary?

src/Latte/Template.php
```diff
@@ -173,13 +173,18 @@ protected function initialize(& $params)
...
+			$this->global->snippetDriver = new SnippetDriver($this->global->snippetBridge);
 		}
 
 		Runtime\Filters::$xhtml = (bool) preg_match('#xml|xhtml#', $this->contentType);
 		// old accumulators for back compatibility
 		$this->params['_l'] = $params['_l'] = new \stdClass;
 		$this->params['_g'] = $params['_g'] = $this->global;
 		$params['_b'] = (object) ['blocks' => & $this->blockQueue, 'types' => & $this->blockTypes];
+		if (isset($this->global->snippetDriver) && $this->global->snippetBridge->isSnippetMode()) {
```
dg: `$this->global->snippetBridge` may not be set

tests/Latte/BlockMacros.snippet.alt.phpt
```diff
@@ -0,0 +1,20 @@
+<?php
+
+/**
+ * Test: general snippets test.
+ */
+
+use Nette\Utils\Html;
+use Nette\Bridges\ApplicationLatte\UIMacros;
```
dg: UIMacros is needed?

dg:
This look really cool, I'm :+1: for merge.

src/Latte/ISnippetBridge.php
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	public function getHtmlId($name);
+
+
+	/**
+	 * @param string snippet name
+	 * @param string html content
+	 * @return mixed
+	 */
+	public function addToPayload($name, $content);
```
dg: `addSnippet`?

src/Latte/SnippetDriver.php
```diff
@@ -0,0 +1,102 @@
...
+
+	/** @var int */
+	protected $renderingLevel = 0;
+
```
dg: $nestingLevel? or deep?

src/Latte/SnippetDriver.php
```diff
@@ -0,0 +1,102 @@
...
+		if (($this->renderingLevel === 0 && $this->bridge->isInvalid($name))
+			|| ($type === self::TYPE_DYNAMIC && ($previous = end($this->stack)) && $previous[1] === TRUE)) {
+			ob_start(function () {});
+			$this->renderingLevel = $type === self::TYPE_AREA ? 0 : 1;
+			$obStarted = TRUE;
+		} elseif ($this->renderingLevel > 0) {
+			$this->renderingLevel++;
+		}
+		$this->stack[] = [$name, $obStarted];
+		if($name !== "") {
```
dg: Isn't it better to check for empty name inside bridge?

src/Latte/Macros/BlockMacros.php
matej21: value would be resolved in the compile time, not sure it is ok..

dg: I know, but it is better to have readable source code that generated source code.

matej21:
Updated 2b99d123044f
Is it ok now? if so, i'll squash it and let's merge! :)

dg:
Great!

I am thinking … when SnippetDrivers is in Latte\Runtime, ISnippetBridge should be in Runtime too, or not? It is used only by Driver, so … probably should.

matej21:
ok, updated

dg:
Merged 3d1a58f75ac4bacbcd2154708e3c6e30596b9641 thanks!

dg:
I don't know why, but this noticeably slows down web page with ublaboo/datagrid

matej21:
In snippet mode or non-snippet mode? or both? will you investigate or should I?

dg:
In non-snippet mode. I tried to use profiler, but with no success.

PR #49, opened 2015-01-18
Engine: optimized loadCacheFile() performance

JanTvrdik:
It's mostly similar to what `DI\ContainerLoader` does. Shouldn't we write the code to temporary file first and them use `rename`?

---

Edit: I've tried slightly different approach. Will it work?

### DIFF
--- a/src/Latte/Engine.php
+++ b/src/Latte/Engine.php
@@ -155,27 +155,43 @@ class Engine extends Object
 
 		$file = $this->getCacheFile($name);
 
-		$handle = fopen($file, 'c+');
-		if (!$handle) {
-			throw new \RuntimeException("Unable to open or create file '$file'.");
-		}
-		flock($handle, LOCK_SH);
-		$stat = fstat($handle);
-		if (!$stat['size'] || ($this->autoRefresh && $this->getLoader()->isExpired($name, $stat['mtime']))) {
-			ftruncate($handle, 0);
-			flock($handle, LOCK_EX);
-			$stat = fstat($handle);
-			if (!$stat['size']) {
-				$code = $this->compile($name);
-				if (fwrite($handle, $code, strlen($code)) !== strlen($code)) {
-					ftruncate($handle, 0);
-					throw new \RuntimeException("Unable to write file '$file'.");
-				}
-			}
-			flock($handle, LOCK_SH); // holds the lock
+		if (!$this->isExpired($file, $name) && (@include $file) !== FALSE) { // @ - file may not exist
+			return;
 		}
 
-		include $file;
+		if (!is_dir($this->tempDirectory)) {
+			@mkdir($this->tempDirectory); // @ - directory may already exist
+		}
+
+		$handle = fopen("$file.lock", 'c+');
+		if (!$handle || !flock($handle, LOCK_EX)) {
+			throw new \RuntimeException("Unable to acquire exclusive lock '$file.lock'.");
+		}
+
+		if (!is_file($file) || $this->isExpired($file, $name)) {
+			$code = $this->compile($name);
+			if (file_put_contents("$file.tmp", $code) !== strlen($code) || !rename("$file.tmp", $file)) {
+				@unlink("$file.tmp"); // @ - file may not exist
+				throw new \RuntimeException("Unable to create '$file'.");
+			}
+		}
+
+		if ((include $file) === FALSE) {
+			throw new \RuntimeException("Unable to load '$file'.");
+		}
+
+		flock($handle, LOCK_UN);
+	}
+
+
+	/**
+	 * @param  string
+	 * @param  string
+	 * @return bool
+	 */
+	private function isExpired($file, $name)
+	{
+		return $this->autoRefresh && $this->getLoader()->isExpired($name, (int) @filemtime($file)); // @ - file may not exist
 	}
 
 
@@ -184,14 +200,6 @@ class Engine extends Object
 	 */
 	public function getCacheFile($name)
 	{
-		if (!$this->tempDirectory) {
-			throw new \RuntimeException('Set path to temporary directory using setTempDirectory().');
-		} elseif (!is_dir($this->tempDirectory)) {
-			@mkdir($this->tempDirectory); // High concurrency
-			if (!is_dir($this->tempDirectory)) {
-				throw new \RuntimeException("Temporary directory cannot be created. Check access rights");
-			}
-		}
 		$file = md5($name);
 		if (preg_match('#\b\w.{10,50}$#', $name, $m)) {
 			$file = trim(preg_replace('#\W+#', '-', $m[0]), '-') . '-' . $file;


JanTvrdik:
I think this should work. Any thoughts?

src/Latte/Engine.php
```diff
@@ -156,33 +156,52 @@ private function loadCacheFile($name, $params)
...
+		}
+	}
+
+
+	/**
+	 * @param  string
+	 * @param  array
+	 * @return bool
+	 */
+	private function tryInclude(/*$file, array $params*/)
```
fprochazka: I think this should be static

dg:
How your approach differs?

src/Latte/Engine.php
JanTvrdik: Good point!

JanTvrdik:
1) It should faster, because in most common use-case (disabled autorefresh and template is already compiled) we directly include the template without using fopen, flock and fstat. See https://github.com/nette/di/commit/7f86308c2618d42282a4ba065f218f62a2f2f82d
2) It writes the compiled template first to a temporary file which is then renamed.

dg:
I mean how it (and why) differs from solution used in DI.

JanTvrdik:
@dg DI does not use atomic write (which requires rename)-

JanTvrdik:
I'm wondering whether it is possible to unlink the `*.lock` files without breaking anything.

src/Latte/Engine.php
```diff
@@ -156,33 +156,53 @@ private function loadCacheFile($name, $params)
...
-			$stat = fstat($handle);
-			if (!$stat['size']) {
-				$code = $this->compile($name);
-				if (fwrite($handle, $code, strlen($code)) !== strlen($code)) {
-					ftruncate($handle, 0);
-					throw new \RuntimeException("Unable to write file '$file'.");
-				}
+
+		if (flock($handle, LOCK_EX) && !is_file($file)) {
+			$code = $this->compile($name);
```
dg: It look good, I'm just not sure if this is flock() should be in condition…

JanTvrdik: `flock` may fail. We should not compile the template if we don't have the lock.

dg: Can you please add comments for shupup operators?

JanTvrdik: It may be even better to throw exception on `flock` failure as it should never fail.

dg: When flock fails, it is relatively good, the only problem is that it will create the same template in multiple threads.

JanTvrdik: Which may in theory damage the compiled file because each of those thread may write something different if source template will change in between compile calls. I think that exception is fine because `flock` should never time out it.

dg: I still see no problem.

JanTvrdik: I tried to make a table what can happen if locks are not used of failure is ignored:

| Thread A          | Thread B          | outside of PHP       |
|-------------------|-------------------|----------------------|
| fopen             |                   |                      |
|                   | fopen             |                      |
| compile           |                   |                      |
|                   |                   | source.latte changed |
|                   | compile           |                      |
| file_put_contents | file_put_contents |                      |

If the compiled template has the same size in both thread A and B, then the it will result in broken template (hybrid from two compilation).

dg: Ok. Can you squash it?

JanTvrdik:
Hm… I discovered an issue in the current implementation – the file may be deleted between `is_file` check and `tryInclude`. It is unlikely, but possible. It could be solved be acquiring the same lock before calling `unlink`.

dg:
Acquire a shared lock before isExpired(), like DI does.

JanTvrdik:
Thinking about thread-safety is killing my brain…

The question is – do we care about performance with *autoRefresh* enabled or not? Because it is a lot simpler if we do not. But I'm afraid that we do. The solution used by DI is not reliable.

JanTvrdik:
Another rewrite and I'm still not satisfied.

JanTvrdik:
Few points:

* You need to check `is_dir($this->tempDirectory)` before any `fopen`
* You need to have at least shared lock from `!is_file($file)` to the final `self::tryInclude` (inclusive)
* You need to have exclusive lock when doing `unlink` and `compile` + `rename`

JanTvrdik:
Note: it would be best to verify the expiration once again after acquiring the exclusive lock but I think it is unnecessary.

src/Latte/Engine.php
```diff
@@ -156,33 +156,62 @@ private function loadCacheFile($name, $params)
...
-				$code = $this->compile($name);
-				if (fwrite($handle, $code, strlen($code)) !== strlen($code)) {
-					ftruncate($handle, 0);
-					throw new \RuntimeException("Unable to write file '$file'.");
-				}
-			}
-			flock($handle, LOCK_SH); // holds the lock
+
+		if (!is_dir($this->tempDirectory)) {
+			@mkdir($this->tempDirectory); // @ - directory may already exist
```
milo: @JanTvrdik What about `if (@!mkdir($this->tempDirectory) && !is_dir($this->tempDirectory)) throw ... 'Unable to create temp dir'`?

milo:
Just idea... what about `E_WARNING` and `eval()` instead of all exception?

src/Latte/Engine.php
JanTvrdik: It is not necessary but yes, we can do that.

JanTvrdik:
@milo No idea what you mean, sorry. Could you explain it a bit better?

milo:
Seems to me, that every exception can occure only when application is under heavy load and someone purges the temp dir or edits the code. The exceptions will lead to `HTTP 500`. To prevent this, generated code can be `eval()`d in the same way as `$this->tempDirectory` is null.

JanTvrdik:
Thread-safety is hard. Really hard. I discovered another issue in the implementation. When autorefresh is enabled, the compiled file may be **unlinked while being included** by another thread which leads in most cases to parse error. It is very unlikely, but possible in theory.

fprochazka:
@JanTvrdik there are possible scenarios and acceptable errors. IMHO focusing on the main use-cases and ignoring the rest would be beneficial.

milo:
@JanTvrdik And why unlink at all? When expired, it will expire in all following threads, so replace is fine, isn't?

JanTvrdik:
@fprochazka Possible parser error is from my point of view unacceptable on production server. I would be however perfectly fine with saying that people should disable autorefresh on production server (which is default behavior) if they need strong thread-safety guarantees.

@milo Yes, not doing unlink at all would probably be the best solution.

milo:
@JanTvrdik OT, have you some measurement how fast is your implementation without original share lock? Or is it surprice? :)

JanTvrdik:
@milo No, I didn't to measure anything. I think it is safe to assume that not using lock is faster. Feel free to measure the speedup, if you have the time.

dg:
> Thinking about thread-safety is killing my brain…

I know very well what you mean :-)

JanTvrdik:
Rebased and squashed on current master.

dg:
:+1:

JanTvrdik:
Lets talk about when the current implementation may fail.

1. **External change or deletion** of compiled template. This will most-often happen when someone deletes `temp` directory on production server. AFAIK there is nothing we can do this.

2. When *autoRefresh* is enabled there is a possible **include-rename collision**. Under very (very!) rare circumstances `rename` may overwrite file while the file is being included by another thread. Real-world simulation of this problem is very close to impossible so I looked into *include-rename collision* in a general in a simulated environment outside of Latte. On Windows it may lead to `rename` failure due to "Access is denied." error, On Linux this collision did not cause any issue.

----

I think it is ready for merge. cc @dg

milo:
@JanTvrdik nice :+1: 

Did you consider my proposal to replace every `\RuntimeException` by `E_USER_WARNING` and `eval($code)`? Overstated?

JanTvrdik:
@milo Yes and I think it is too dangerous (with wrong permission you will recompile and eval the code on every request) and even if we choose to do it, there are other problems with deleting `temp` on production which can not be avoided, such as parse errors.

---

One more thing – I removed `@`  before the final `include` as it expected to never fail. We may even consider replaced the entire condition with simple `require $file`.

dg:
Thank you!

PR #50, opened 2015-01-29
Engine: inlined constant values to avoid loading Compiler in runtime

JanTvrdik:
Also the Compiler also has `CONTENT_URL`. Should we add it? Should the `Compiler` be updated to load the constant values from the `Engine`?


dg:
Probably yes­…

JanTvrdik:
done

PR #52, opened 2015-01-29
Filters: few BC breaks to consider

JanTvrdik:


fprochazka:
Case sensitivity is a nasty one. However, it can be really easily fixed.

dg:
The need to receive arguments as array will disappear with PHP 5.6, so it is imho unnecessary.

JanTvrdik:
Yes, with PHP 5.6+ it is a lot easier to write "filter loader". The proposed solution makes it easy to write "filter loader" in older versions and removes the useless `array_unshift`.

dg:
Case sensitivity for Latte filters is really nasty, BC break for dynamic filters too.

PR #53, opened 2015-01-30
BlockMacros: runtime helpers moved to BlockMacrosRuntime

JanTvrdik:
Splitting prevents loading `MacroSet.php` and `IMacro.php` files in runtime. Were the runtime helpers considered internal or should I keep them in `BlockMacros` for compatibility?


dg:
:+1:

PR #57, opened 2015-02-12
Parser: fixed syntax=off

matej21:
fixes http://forum.nette.org/cs/22145-makro-n-syntax-off-sa-vypne-skor-ako-by-malo
#drunkPR

### DIFF
--- a/src/Latte/Parser.php
+++ b/src/Latte/Parser.php
@@ -57,6 +57,9 @@ class Parser extends Object
 	/** @var string used by filter() */
 	private $syntaxEndTag;
 
+	/** @var int */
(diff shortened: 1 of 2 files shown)


dg:
Heh, I just fixed it too ;)

matej21:
I was first! :D

dg:
Not only the first, but your solution is better than my.

dg:
Please change Lvl to Level, short names are démodé https://github.com/nette/bootstrap/commit/6ef8507bccdd3ebe0fdd2aefed17a3ea53015a62

matej21:
fixed

dg:
Thx!

PR #60, opened 2015-02-27
Failing test: not escaped included block in meta tag

stekycz:
It is very simple to use Latte incorrectly without correct escape.


matej21:
I tried to solve it this way: matej21/latte@366fc515d44 what do you think?

dg:
Fixed

PR #61, opened 2015-03-08
Compiler: <script> with type text/json is escaped as javascript

kukulich:


JanTvrdik:
JSON mime type is `application/json`. Why do you need `text/json`?

kukulich:
Force pushed with `application/json`, however it doesn't matter if there's `application/json` or `text/json`. Both work with this pull request.

dg:
Can you prepend commit with `Compiler:`, ie. `Compiler: text/json is escaped as javascript`?

kukulich:
@dg Done

dg:
Thanks

PR #65, opened 2015-04-15
Include block: trigger warning on incompatible context

matej21:
ref #60

### DIFF
--- a/src/Latte/Macros/BlockMacros.php
+++ b/src/Latte/Macros/BlockMacros.php
@@ -21,6 +21,9 @@ class BlockMacros extends MacroSet
 	/** @var array */
 	private $namedBlocks = [];
 
+	/** @var array */
+	private $blockTypes = [];
+
 	/** @var bool */
 	private $extends;
 
@@ -48,6 +51,7 @@ class BlockMacros extends MacroSet
 	public function initialize()
 	{
 		$this->namedBlocks = [];
+		$this->blockTypes = [];
 		$this->extends = NULL;
 	}
 
@@ -72,6 +76,7 @@ class BlockMacros extends MacroSet
 				$this->getCompiler()->addMethod($functions[$name], $code, '$_b, $_args');
 			}
 			$this->getCompiler()->addProperty('_blocks', $functions);
+			$this->getCompiler()->addProperty('_blockTypes', $this->blockTypes);
 		}
 
 
@@ -121,6 +126,13 @@ class BlockMacros extends MacroSet
 			$cmd = "call_user_func(reset(\$_b->blocks[$name]), \$_b, %node.array? + get_defined_vars())";
 		} else {
 			$cmd = 'Latte\Macros\BlockMacrosRuntime::callBlock' . ($parent ? 'Parent' : '') . "(\$_b, $name, %node.array? + " . ($parent ? 'get_defined_vars()' : '$this->params') . ')';
+		}
+
+		$node->modifiers = preg_replace('#\|noescape\s?(?=\||\z)#i', '', $node->modifiers, -1, $found);
+		if (!$found && !preg_match('#\|?escape(?:html|htmlcomment|ical|js|url|xml)?\s?(?=\||\z)#i', $node->modifiers)) {
+			$cmd = "if (" . var_export($this->exportBlockType(), TRUE) . " !== \$_b->types[$name]) { "
+				. "trigger_error('Incompatible context for including block $destination.', E_USER_WARNING); }\n"
+				. $cmd;
 		}
 
 		if ($node->modifiers) {
@@ -217,7 +229,9 @@ class BlockMacros extends MacroSet
 				$node->data->func = 'block_' . preg_replace('#[^a-z0-9_]#i', '_', $name) . '_' . substr(md5($name), 0, 7);
 				$fname = $writer->formatWord($name);
 				$node->closingCode = '<?php ' . ($node->name === 'define' ? '' : "call_user_func(reset(\$_b->blocks[$fname]), \$_b, get_defined_vars())") . ' ?>';
-				return "\$_b->blocks[$fname][] = [\$this, '{$node->data->func}'];";
+				$blockType = var_export($this->exportBlockType(), TRUE);
+				return "Latte\\Macros\\BlockMacrosRuntime::checkType($blockType, \$_b->types, $fname);"
+					. "\$_b->blocks[$fname][] = [\$this, '{$node->data->func}'];";
 			}
 		}
 
@@ -235,6 +249,7 @@ class BlockMacros extends MacroSet
 
 		$prolog = $this->namedBlocks ? '' : "if (\$_l->extends) { ob_end_clean(); return \$this->renderChildTemplate(\$_l->extends, get_defined_vars()); }\n";
 		$this->namedBlocks[$name] = TRUE;
+		$this->blockTypes[$name] = $this->exportBlockType();
 
 		$include = 'call_user_func(reset($_b->blocks[%var]), $_b, ' . (($node->name === 'snippet' || $node->name === 'snippetArea') ? '$this->params' : 'get_defined_vars()') . ')';
 		if ($node->modifiers) {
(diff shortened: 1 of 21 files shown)


dg:
:+1:

tests/Latte/macros.include.escape.phpt
```diff
@@ -0,0 +1,40 @@
...
+
+
+$latte = new Latte\Engine;
+$latte->setTempDirectory(TEMP_DIR);
+$latte->setLoader(new \Latte\Loaders\StringLoader());
+
+Assert::error(function () use ($latte) {
+	$latte->renderToString('<meta content="{include #foo}">{block #foo}{$value}{/block}', array('value' => 'b"ar'));
+}, E_USER_WARNING, 'Incompatible context for include, use either |escape or |noescape modifier');
+
```
dg: I think that "use either |escape or |noescape modifier" is misleading.

matej21: any better idea?

src/Latte/Macros/BlockMacros.php
```diff
@@ -125,6 +129,13 @@ public function macroInclude(MacroNode $node, PhpWriter $writer)
 		} else {
 			$cmd = 'Latte\Macros\BlockMacrosRuntime::callBlock' . ($parent ? 'Parent' : '') . "(\$_b, $name, %node.array? + " . ($parent ? 'get_defined_vars' : '$template->getParameters') . '())';
 		}
+		$context = $this->getCompiler()->getContext() ? : array(NULL, NULL);
```
matej21: I'm not sure about this. Is there any difference between `getContext() = NULL` and `getContext() = array(NULL, NULL)`?

tests/Latte/macros.include.escape.phpt
dg: `'Incompatible context for include'` seems fine.

matej21: I think it should say how to fix that.

dg: How to fix it or how to bypass it?

matej21: I don't think it as a bypass. Latte just does not know how to handle it properly, so it leaves the decision to the developer.

matej21:
updated, squashed. I think it is ready to merge.

dg:
IMHO `{block double double}` is not very intuitive, what about `{block "double"}?

matej21:
that is not very explicit. usually, you don't name a block `double` but sth like `{block foo double}`  :)

dg:
Of course something like `{block "foo"}`

matej21:
I don't agree. I think we should use explicit syntax. Your version does not explicitly indicate that it somehow change the behaviour. It just seems like a quoted string. And some other developer unaware of this behaviour may remove it thinking that it is useless.

dg:
What do you don't agree with? I am not suggesting that `{block "foo"}` is good solution, it is just an idea. My point is that relation between word `double` and HTML quotes is not predictable.

matej21:
This feature will not be used very often so I think we should be as explicit as possible. Maybe `{block foo doubleQuotes}`?

dg:
This is little bit complicated… because context is not only about escaping, it is about parsing too. The workaround is

```
<foo bar="{block foo}…{/block}">
```

src/Latte/Macros/BlockMacros.php
```diff
@@ -71,6 +74,7 @@ public function finalize()
 				$func = '_lb' . substr(md5($this->getCompiler()->getTemplateId() . $name), 0, 10) . '_' . preg_replace('#[^a-z0-9_]#i', '_', $name);
 				$snippet = $name[0] === '_';
 				$prolog[] = "//\n// block $name\n//\n"
+					. "\$_b->types[" . var_export($name, TRUE) . "][] = " . var_export($this->blockTypes[$name], TRUE) . ";"
```
dg: The `[]` means, that blocks with same name can have a different contexts? Is it reasonable?

dg:
Btw, I think we can always escape both quotes and remove difference between single / double / no quotes.

src/Latte/Macros/BlockMacros.php
matej21: It just reflects blocks, which also use `[]`. I'm not really sure why :)

dg: Blocks have more definitions (more functions), but IMHO it should have only one type.

matej21: Ok, it is probably needless. I will update it and will try do find some failing test :)

matej21: Cannot be there any issues with dynamic blocks and/or overriding block from a parent template?

dg: Overriding block must have the same context as overrides, I think there is no issue…

matej21:
rebased and partially rewritten. ad block overriding - there actually was an issue, but it should be fixed now :)

I've removed that second commit. Your solution/workaround is better. And in the future we can add sth like `n:omit` which will just extract a block and omit surrounding HTML.

dg:
Thanks, I'll merge it today

matej21:
Ok, I'll rebase once again (in an hour or so) and will update commit message and test names. Yesterday I was too focused on the functionality :)

matej21:
Well.. with recent changes it has to be rewritten (almost) completely.

dg:
What about something like this? https://github.com/dg/latte/commits/block-type

matej21:
@dg that looks great!

dg:
Feel free to take over and continue

matej21:
rebased, renamed tests

src/Latte/Macros/BlockMacros.php
```diff
@@ -123,6 +127,13 @@ public function macroInclude(MacroNode $node, PhpWriter $writer)
 			$cmd = 'Latte\Macros\BlockMacrosRuntime::callBlock' . ($parent ? 'Parent' : '') . "(\$_b, $name, %node.array? + " . ($parent ? 'get_defined_vars()' : '$this->params') . ')';
 		}
 
+		$node->modifiers = preg_replace('#\|noescape\s?(?=\||\z)#i', '', $node->modifiers, -1, $found);
+		if (!$found && !preg_match('#\|?escape(?:html|htmlcomment|ical|js|url|xml)?\s?(?=\||\z)#i', $node->modifiers)) {
+			$cmd = "if (" . var_export($this->exportBlockType(), TRUE) . " !== \$_b->types[$name]) { "
+				. "trigger_error('Incompatible context for include block $destination.', E_USER_WARNING); }\n"
```
dg: This is my fault, but maybe `for including` or `to include` is better.

dg:
There is the last thing I am not sure. `|noescape` is here used as `|nocontextcheck`, right? Isn't it confusing? I removed it from exception message because it is not good idea to advise people to use `noescape` at the moment they do not know what is happening.

matej21:
> There is the last thing I am not sure. `|noescape` is here used as `|nocontextcheck`, right? 

yes, that is right. 

> I removed it from exception message...

maybe we can write there sth like "escape it or define it in proper context" ?

dg:
"define it in proper context" is good, but little bit obvious, i.e. "exception: block is in incompatible context" and "exception: block is in incompatible context, define it in proper context" give the same information.

"escape it" what do you mean?

dg:
what about `|nowarning`? I'll can use it in `{foreach}` too.

matej21:
By "escape" I mean "use `|escape` filter"..

noescape was really misleading, nocheck is ok, but nowarning was imho better :)

dg:
I thinking about use nocheck as alias for (very bad name) nosafeurl, thus nocheck is better than nowarning.

matej21:
ok, I will use |nocheck. What about the error message? should I include fixing possibilities (escape, nocheck) or the current message is enough?

dg:
How to solve context incompatibility is really complex issue, so IMHO it will be better to use simple exception message and let user to google  doc/forum…

matej21:
updated and added missing https://github.com/nette/latte/pull/65/files#diff-232e670f239e959d76e63b24a5bc9cb2R54

dg:
Great! Can I merge it?

matej21:
I think you can.

matej21:
I'm an idiot. I forgot to change noescape to nocheck :)

dg:
I'll fix it

dg:
There is problem. For `<h1 n:block=title>` is context `html, tag`, because block is defined inside tag.

matej21:
ok. should I send PR?

dg:
Done

matej21:
You should also check script/style tag, which has different content type: https://github.com/nette/latte/blob/master/src%2FLatte%2FCompiler.php#L359-L363

dg:
Uff, it is really hard

PR #69, opened 2015-05-28
Update nl2br to return Latte\Runtime\Html object to prevent a need for …

finwe:
- feature
- issues - none
- documentation - not needed
- BC break - possible

nl2br helper returns Html object instead of string. This prevents need to use `!` in `{!$foo|nl2br}` or abominations as `{$foo|escape|nl2br|noescape}`.

BC break is possible on edge use cases, but unlikely as this filter is mainly used to output the contents as a string and `Latte\Runtime\Html` object has `__toString` magic method implemented.

### DIFF
--- a/src/Latte/Runtime/Filters.php
+++ b/src/Latte/Runtime/Filters.php
@@ -270,7 +270,7 @@ class Filters
 	 */
 	public static function nl2br($value)
 	{
-		return nl2br($value, self::$xhtml);
+		return new Html(nl2br($value, self::$xhtml));
 	}
 
 
--- a/tests/Latte/Filters.nl2br().phpt
+++ b/tests/Latte/Filters.nl2br().phpt
@@ -14,7 +14,7 @@ require __DIR__ . '/../bootstrap.php';
 $input = "Hello\nmy\r\nfriend\n\r";
 
 Filters::$xhtml = TRUE;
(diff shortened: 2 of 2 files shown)


finwe:
OK, I guess I'll push force with Latte\Runtime\Html

dg:
Can you change message to format "Filters: ...."?

finwe:
Done.

dg:
Thx

Aurielle:
This is a massive security issue and also something that goes directly against the principle of Latte, since it outputs non-escaped string with *no visible information about it*. We had `!` or `|noescape` to see what comes out unescaped, this just says nothing.

This change also prevents me from doing anything else with the string that I'd want to output in raw format (e.g. replacing URLs with clickable links), since it just escapes the rest of the string. I suggest a new modifier with a different name and reverting nl2br to the way it was before.

JanTvrdik:
> This change also prevents me from doing anything

Why? Both `{$text|nl2br|links2url}` and {$text|links2url|nl2br} should work just fine. The implementation may look like this:

~~~php
function links2url($value)
{
    $value = self::escapeHtml($value); 
    $value = Strings::replace($value, ..., ...); // convert URLs to <a>
    return new Html($value); // mark the string as HTML.
}
~~~

PR #70, opened 2015-06-02
Revert Filters: Update nl2br to return Latte\Runtime\Html object

finwe:
Former change could cause potential hidden XSS errors as the whole string was being unescaped.

This reverts commit 86b55f5d8e7b0ed182404df987107e905a7f974f.

### DIFF
--- a/src/Latte/Runtime/Filters.php
+++ b/src/Latte/Runtime/Filters.php
@@ -270,7 +270,7 @@ class Filters
 	 */
 	public static function nl2br($value)
 	{
-		return new Html(nl2br($value, self::$xhtml));
+		return nl2br($value, self::$xhtml);
 	}
 
 
--- a/tests/Latte/Filters.nl2br().phpt
+++ b/tests/Latte/Filters.nl2br().phpt
@@ -14,7 +14,7 @@ require __DIR__ . '/../bootstrap.php';
 $input = "Hello\nmy\r\nfriend\n\r";
 
 Filters::$xhtml = TRUE;
-Assert::same( "Hello<br />\nmy<br />\r\nfriend<br />\n\r", (string) Filters::nl2br($input) );
+Assert::same( "Hello<br />\nmy<br />\r\nfriend<br />\n\r", Filters::nl2br($input) );
 
 Filters::$xhtml = FALSE;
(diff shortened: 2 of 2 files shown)


JanTvrdik:
Why was this solution preferred over fixing the implementation?

finwe:
Quicker, the most safe, return to the safe state of things.

If you have a safe implementation, you are welcome to submit it as a PR yourself, instead of bitchin' around about others' solutions.

fprochazka:
I don't think that pointing out security issues is bitching.

finwe:
Sure, by itself it is not.

dg:
„Why was this solution preferred over fixing the implementation?“ is definitely bitching.

JanTvrdik:
How is trying to figure out why something was done certain way bitching? It was not meant as "you are idiots for choosing sth" but as "i don't understand why did you choose a solution which provide worse UX". Was there a problem with the proposed solution? Was the revert chosen because it was the fastest and most straightforward way to fix a security issue? I don't know. So I asked.

finwe:
Oh, when did you change the original comment to include the proposed solution? Sorry, there was no notification about it so I missed it entirely.

Following conversation was therefore "a bit" misunderstood.

I cannot properly assess possible uses and caveats of the solution, so from my point of wiew the revert is the fastest and most straightforward way to fix a security issue (introduced by me), as you have written yourself.

dg:
@JanTvrdik „proposed solution“ is what?

finwe:
I believe it is

```php
return new Html(nl2br(self::escapeHtml($value), self::$xhtml));
```

which was added to [the original comment](https://github.com/nette/latte/commit/86b55f5d8e7b0ed182404df987107e905a7f974f#commitcomment-11424716) by edit some time later.

dg:
It is BC break.

JanTvrdik:
Because `{$text|escapeHtml|nl2br|noescape}` will no longer work?
Is it acceptable BC break for Latte 2.4?
What about changing `escapeHtml` filter (when used explicitly) to return `Html` object?

dg:
There are other use cases:

```
		{block|nl2br}
			{$a}
			{$b}
		{/block}
```

JanTvrdik:
Good point! Another random idea: We may introduce `html` (or `safeHtml`?) helper which would wrap string as `Html` object. You could then write `{block|html|nl2br}`.

PR #72, opened 2015-06-19
Method loadCacheFile() is public

ZZromanZZ:
Makes simple way to compile Latte files globally:

```php
foreach (\Nette\Utils\Finder::findFiles("*.latte")->from(APP_DIR) as $file) {
     $latteEngine->createCache($file->getRealPath());
}
```


lookyman:
IMHO this can be closed thanks to 29929e57db3e91b72dbf84b2ea5453f9f1644343. Ping @dg.

PR #73, opened 2015-06-28
BlockMacros: fixed compatibility with new substr behavior on PHP 7-alpha2

JanTvrdik:
https://github.com/php/php-src/commit/257054e81d4dad73bf9d09cd206d3a6727ad1777
http://3v4l.org/uoon2

Fixed tests in nette/application. Not sure what is the best fix because I don't understand snippets.


JanTvrdik:
@dg Was passing `FALSE` in first `redrawControl` argument (which is later cast to `0`) intended?

dg:
I have no idea how it works :-(

dg:
Is it still actual?

JanTvrdik:
AFAIK it is already fixed with https://github.com/nette/latte/commit/173b553c9d2747a464fc337138625a957f9eae39 and https://github.com/nette/application/commit/1101be47c6b251b05d2d2e7523ae9390ab2ad016.

Issue #55, opened 2015-01-31
Using Latte in read-only (pre-generated) mode


Majkl578:
It't not really feasible to use Latte on read-only filesystem or in an environment where PHP files should not be generated/modified. This includes for example:
- HHVM in [RepoAuthoritative mode](https://github.com/facebook/hhvm/wiki/Performance-Tuning) - any generated files are essentially forbidden, since whole code base is compiled into bytecode (and thus could be distributed entirely as a single hhbc file without source files themselves);
- Google App Engine - file system is strictly read-only

This is something that's bugging me for a while (ping @matej21), and after Latte separation, it is even worse because cache file generation is hardcoded in Engine.

Basically what I'd expect is some way to pre-generate templates and let them be used on production without need to be (re)compiled.
This is also something that could be beneficial for standard deployment as whole checking and recompiling logic could be avoided (I'm looking at you, loadCacheFile).

JanTvrdik:
Send PR.

Majkl578:
:)
I'd like to discuss the best approach first. In 2.1, it was somehow separated by using nette/caching, but it was removed and hardcoded into Engine. So, what do you suggest?
- Separate caching and loading from Engine (it violates SRP anyway)?
- Somehow hack Engine to support read-only mode?
- ??? (suggest other way)

JanTvrdik:
Well I thought that a simple method `Engine::ensureThatTemplateIsCompiled($template)` would be enough.

mishak87:
@Majkl578 Where exactly is the problem?

I assume you are running artefact in known environment made from source code?
If so you can pre/generate all templates with the knowledge of target environments tempDirectory and hash it will produce for [each template](https://github.com/nette/latte/blob/v2.3/src/Latte/Engine.php#L223).

Last resort would be to set temp directory to simple value (`null` or `latte` etc.) and load them [before calling render](https://github.com/nette/latte/blob/v2.3/src/Latte/Engine.php#L91).

PR #74, opened 2015-07-04
Engine: warmupCache()

pkristian:
- feature
- issues - #55
- BC break - no

Caching named lattes without rendering.

usage:
 ```php
$latte = new Latte\Engine;
$latte->setTempDirectory('/myCache');

$latte->warmupCache('something.latte');
```

note: Temp Directory must be set for point of this functionality

### DIFF
--- a/src/Latte/Engine.php
+++ b/src/Latte/Engine.php
@@ -149,6 +149,25 @@ class Engine extends Object
 
 
 	/**
+	 * Puts specific latte to cache
+	 * @param string $name
+	 *
+	 * @throws \Latte\RuntimeException
+	 */
+	public function warmupCache($name)
+	{
+		if (!isset($this->tempDirectory)) {
+			throw new RuntimeException('Temp directory is not set, so putting to cache is useless.');
+		}
+
+		$class = $this->getTemplateClass($name);
+		if (!class_exists($class, FALSE)) {
+			$this->loadCacheFile($name);
+		}
+	}
+
+
+	/**
 	 * @return void
 	 */
 	private function loadCacheFile($name)
--- a/tests/Latte/Engine.warmupCache.phpt
+++ b/tests/Latte/Engine.warmupCache.phpt
@@ -1,0 +1,25 @@
+<?php
+
+/**
+ * Test: Latte\Engine: {warmupCache}
+ */
+
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+test(function () {
+	$precachedDir =__DIR__.'/../tmp/';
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/Latte/Engine.php [hunk 1]
```diff
@@ -149,6 +149,29 @@ public function compile($name)
 
 
 	/**
+	 * Puts specific latte to cache
+	 * @param $name
+	 *
+	 * @throws \Latte\RuntimeException
+	 */
+	public function putToCache($name)
```
fprochazka: `warmupCache` might be a better name

src/Latte/Engine.php
```diff
@@ -149,6 +149,29 @@ public function compile($name)
...
+	{
+		if (!isset($this->tempDirectory))
+		{
```
fprochazka: cs?

tests/Latte/Engine.putToCache.phpt
```diff
@@ -0,0 +1,31 @@
...
+
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+test(function () {
+
+	$precachedDir =__DIR__.'/../tmp/';
+	$template = __DIR__.'/templates/general.latte';
```
fprochazka: also cs

src/Latte/Engine.php
```diff
@@ -149,6 +149,27 @@ public function compile($name)
 
 
 	/**
+	 * Puts specific latte to cache
+	 * @param $name
```
fprochazka: `@param string`

src/Latte/Engine.php
```diff
@@ -149,6 +149,27 @@ public function compile($name)
...
+	 *
+	 * @throws \Latte\RuntimeException
+	 */
+	public function warmupCache($name)
+	{
+		if (!isset($this->tempDirectory)) {
+			throw new RuntimeException('Temp directory is not set, so putting to cache is useless.');
+		}
+
+		$this->setAutoRefresh(FALSE);
```
lookyman: Doesn't this need to get turned back on after, if it was on before?

pkristian: I don't think so. If you are creating this, you do it in clean instance, with another directory, or not?

JanTvrdik: Why do you need to set autorefresh to false?

pkristian: Don't need perhabs

dg:
Reverted… 

Use `file_put_contents($latte->getCacheFile($name), $latte->compile($name));`

PR #75, opened 2015-08-07
Translation

kostaldavid8:
I missed translations, so I ported it from Nette/Utils.

Added method addTranslator to Latte\Engine (needs instance of
Latte\Runtime\ITranslator)
Added test

- feature
- documentation - needed: update for Latte Macros
- BC break - no


JanTvrdik:
:-1: It just duplicates code.

kostaldavid8:
Yes it uses code from Nette, but method addTranslator in Template is not even implemented in Latte

dg:
This is not needed, you can add any translator via addFilter.

PR #79, opened 2015-09-03
Parser: Fix php error when empty input

juniwalk:
When template file is not in valid UTF-8 format, PHP warning is raised:
```
substr_count(): Length value 1 exceeds string length
```
This PR fixes this warning and let's Latte throw `CompileException` as it is expected.


juniwalk:
I see you had better fix. Thanks for fixing this.

PR #81, opened 2015-09-14
BlockMacros: fixed child template without block [fixes #80]

### LINKED ISSUE #80: Variable existence in extended layout depends on block usage
Following example raises notice about undefined variable ```$var```:

```latte
// view.latte
{extends './layout.latte'}
{var $var = 'ok'}
 

// layout.latte
{$var} // undefined variable
```

but this works:

```latte
// view.latte
{extends './layout.latte'}
{var $var = 'ok'}
{block #test}
 

// layout.latte
{$var}
{include #test}
```

This is very confusing behaviour. Is it possible to fix it?

matej21:


dg:
Thanks

PR #82, opened 2015-09-23
Subresource Integrity: added n:integrity

dg:
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

### DIFF
--- a/src/Latte/Macros/CoreMacros.php
+++ b/src/Latte/Macros/CoreMacros.php
@@ -81,6 +81,7 @@ class CoreMacros extends MacroSet
 
 		$me->addMacro('class', NULL, NULL, [$me, 'macroClass']);
 		$me->addMacro('attr', NULL, NULL, [$me, 'macroAttr']);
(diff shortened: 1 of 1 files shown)


vojtech-dobes:
What if content of the file gets fetched when it's already compromised? Integrity than won't complain, which is wrong - at least if I understand Subresource integrity and this PR correctly.

Unlink:
Maybe it can have optional parameter with path or url to file to calculate hash.
So i can keep verified content locally to calculate hash (or use another CDN for calculation) and serve it from CDN.

f3l1x:
:+1:

lookyman:
"I'm gonna verify the integrity of this downloaded file by.. downloading it? Sure, that will work great!".

As @vojtech-dobes said, isn't this completely broken?

fprochazka:
I don't like it either, you should be able to provide filepath in local filesystem, or stream to read from.

dg:
IMHO attribute `integrity` should be part of documentation like here http://www.bootstrapcdn.com/ (click on arrow).

PR #84, opened 2015-10-04
Latte\Object replaced with trait 

dg:


f3l1x:
:+1:

enumag:
:+1: Can we have that trait in Nette/Utils as well?

PR #85, opened 2015-10-09
Macros: allowed to specify implicit behavior as "inner" or "tag"

richard-ejem:
example:

```
$macroSet->addMacro('alwaysIn', '?>{<?php', '?>}<?php', NULL, MacroNode::PREFIX_INNER);
```

will transform

```
<div n:alwaysIn>foo</div>
```

into

```
<div>{foo}</div>
```


dg:
:+1:

richard-ejem:
just realized a little BC break - classes extending MacroSet and overriding addMacro method will have incompatible method signature.
Can be solved by some magic, like adding addImplicitlyInnerMacro() and addImplicitlyTagMacro() methods, what do you think would be better - this or current solution?

richard-ejem:
I humbly request, can this be merged / some feedback given? :)

dg:
@richard-ejem I want to get it as soon as possible.

dg:
Finally I get to it and it's great!

I am thinking that probably it would not be possible to add macros with the same name and with a different implicit prefix. So implicit prefix can be bounded to macro name instead of pair name & object, what do you think?

dg:
Now I realised sometimes is needed macro that is both inner & outer, like `n:ifcontent`…

richard-ejem:
> it would not be possible to add macros with the same name and with a different implicit prefix
I considered this to be edge case (may be enough to mention in doc), I cannot figure out any reasonable example when that would happen (I also admit that I don't even understand much when it is useful to have any two macros with the same name) - do you have an idea when it could happen? Would be easier to figure out with an example.

> Now I realised sometimes is needed macro that is both inner & outer, like n:ifcontent
Yes, but I would consider this as a separate feature to to implement and discuss (also because it is quite rare case, I think currently used just by ifcontent)

dg:
BTW in your use case, do you need implicit prefixes only for `inner`, or do you need `tag` too?

dg:
Related: https://github.com/nette/latte/commit/79d39dc28786823a92310dd1fa479b65794125bb a usage https://github.com/nette/latte/commit/3e77454fdad297bb1161d9593209993ccfa3341c

richard-ejem:
> BTW in your use case, do you need implicit prefixes only for inner, or do you need tag too?

I needed just inner, but I implemented also tag as it was simple to add and analogical to use.

> Related: 233607a and its usage d38fac5

OK, so both inner/outer macro is solved. I just rebased the stuff, now with flags it is simple and straightforward to bind implicit prefix to macro name, so I fixed it during the rebase.

dg:
Oh, thanks, I keep rebased version in my local repo, I just didn't wrote it.

I mentioned https://github.com/nette/latte/commit/d38fac5fa5266774f8227eaceac2ef13d6adb45a because it maybe solves the same problem. 

Hey, I need you to describe use cases you want to solve :)

richard-ejem:
> Hey, I need you to describe use cases you want to solve :)

For me it's now just one use case - in instante/cms (still in early development) we have sections in page that are editable by client, and they are simply put into template this way:

<div class="..." n:icms.text="homepage-welcome">

</div>

icms.text behaves always as inner (or may be used as common {icms.text} macro when no wrapping element is used).

dg:
Now you have access to inner content via [$innerConent](https://github.com/nette/latte/commit/233607a5cf5dd7a4a568e978a95f8831a203695b#diff-740dcf0d8403fdb55eda81c022c3ce86R61) and you can change it's content [independently](https://github.com/nette/latte/blob/d38fac5fa5266774f8227eaceac2ef13d6adb45a/src/Latte/Macros/CoreMacros.php#L179-L184).

Maybe this is solution to your use case (or maybe not).

The advantage of this solution is that you still have all benefits of outer element, ie. openingCode, closingCode etc.

richard-ejem:
Oh thanks, you are probably true, this is even better solution (so it's upon you if you want to merge implicit prefixes as well or not, for me this should be enough)

dg:
Leave it open and we'll see.

dg:
Latte 3 will have a completely new parser, where it will be handled differently.

PR #88, opened 2015-10-17
Latte: Support for heredoc and nowdoc

sallyx:
Added support for heredoc and nowdoc inside macros
BC break – no


dg:
I think that "Latte-way" is to use `{capture}`.

sallyx:
Heredoc can by used  also with {=}  {?} and {php} tags.  Why not to support it? There could be big chunks of PHP code in CMS build on latte ...

milo:
Aggreed with @dg this is not latte-way.

PR #95, opened 2016-01-15
BlockMacros: fixed enabling snippetMode in the dynamic snippetArea

matej21:
https://forum.nette.org/en/25272-invalidating-dynamic-snippets-with-include-template-is-broken

tests in the nette/application#109


Renat123:
+1

PR #96, opened 2016-01-22
FileLoader: error message explaining touching, when touch() fails

milo:
I was googling in my brain quite long to understand the reason, why is touching done.

### DIFF
--- a/src/Latte/Loaders/FileLoader.php
+++ b/src/Latte/Loaders/FileLoader.php
@@ -27,7 +27,9 @@ class FileLoader implements Latte\ILoader
 			throw new \RuntimeException("Missing template file '$file'.");
 
 		} elseif ($this->isExpired($file, time())) {
-			touch($file);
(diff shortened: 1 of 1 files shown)


dg:
Thanks

src/Latte/Loaders/FileLoader.php
```diff
@@ -27,7 +27,9 @@ public function getContent($file)
 			throw new \RuntimeException("Missing template file '$file'.");
 
 		} elseif ($this->isExpired($file, time())) {
-			touch($file);
+			if (@touch($file) === FALSE) {
+				trigger_error("File's modification time is in the future. Cannot update it: " . error_get_last()['message'], E_USER_WARNING);
```
dg: Is it sure that file's modification time is in the future?

milo: @dg I hope so. The touch() is called only when `isExpired()`, which in this case is compared to current `time()`.

Issue #97, opened 2016-01-22
HTML Attribute names should not be quoted


talyian:
When Latte generates HTML attribute names, they are currently escaped by quoting. This violates the HTML spec since attribute names cannot contain quotes. For example, when rendering the following template with data `["attr" => "a", "value" => "b"]`, 

    <tag {{$attr}}={{$value}} {{$attr}/>

Latte currently generates:

    <tag "a"="b" "a"/>

Instead, the properly formed HTML should be

    <tag a="b" a/>

To prevent adding an XSS vector, only valid HTML Attribute Names should be rendered.

dg:
closed by #98

PR #98, opened 2016-01-22
adds CONTEXT_ATTRNAME status, allows unquoted attribute names

talyian:
This change implements #97 

An attribute-name context is added Compiler.php - it is used in some of the places CONTEXT_UNQUOTED_ATTR was used before. In this context, values are sanitized and output without quotes.

A change to Parser.php is needed to allow picking up attribute values after a macro for the attr name, for example, when parsing `<tag {{$attr}}={{$value}} /> `

Finally, a new filter, escapeName, is added that sanitizes incoming attribute names. It outputs "invalid" instead of `""` for  invalid attribute names since generating `<tag ="foo" />` would also be invalid output. (This could be changed. I'm not set on that since at least there's no XSS in either case)


dg:
Thank you for PR and sorry for late response. I'll tried to solve it in little bit different way.

PR #99, opened 2016-02-01
Engine: fixed CompileException sourceLine on PHP7

matej21:
it is ok in the master, but in v2.3 branch, `$e` is overwritten and wrong line number is passed to the CompileException.


dg:
Thanks

Issue #33, opened 2014-09-01
Feature request: Add helper `|count` for counting things


rostenkowski:
Currently it is possible to use this:
```html
{count($foo)}
```
It would be nice to have common helper for this:
```html
{$foo|count}
```
Maybe all PHP functions could be used as a helper...

milo:
Solved by https://github.com/nette/latte/pull/103

PR #103, opened 2016-02-14
Expose "length" filter, make it work with Countable objects

rundef:
I needed to count the number of elements in an array ...
I saw that this feature has been talked about in issue #33

I wasn't sure if I should have created a new filter `|count` for objects only ... Tell me what you think!

### DIFF
--- a/src/Latte/Runtime/Filters.php
+++ b/src/Latte/Runtime/Filters.php
@@ -366,13 +366,13 @@ class Filters
 	 */
 	public static function length($s)
 	{
-	    if (is_array($s) || $s instanceof \Countable) {
-	        return count($s);
-	    } elseif ($s instanceof \Traversable) {
-	        return iterator_count($s);
-	    } else {
-	        return strlen(utf8_decode($s)); // fastest way
-	    }
+		if (is_array($s) || $s instanceof \Countable) {
+			return count($s);
+		} elseif ($s instanceof \Traversable) {
+			return iterator_count($s);
+		} else {
+			return strlen(utf8_decode($s)); // fastest way
+		}
 	}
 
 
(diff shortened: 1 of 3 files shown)


src/Latte/Runtime/Filters.php
```diff
@@ -366,7 +366,9 @@ public static function capitalize($s)
 	 */
 	public static function length($s)
 	{
-		return strlen(utf8_decode($s)); // fastest way
+		return is_string($s) ?
```
fprochazka: imho it should be something like

```php
public static function length($s)
{
    if (is_array($s)) {
        return count($s);
    } elseif ($s instanceof \Traversable) {
        return iterator_count($s);
    } else {
        return strlen(utf8_decode($s)): // fastest way
    }
```

xificurk:
:+1: 

A test would be nice ;-)

src/Latte/Runtime/Filters.php
dg: if (is_array($val) || $val instanceof \Countable)

rundef:
Thanks for the quick feedback :)

dg:
Thanks, merged

src/Latte/Runtime/Filters.php
```diff
@@ -366,7 +366,13 @@ public static function capitalize($s)
 	 */
 	public static function length($s)
 	{
-		return strlen(utf8_decode($s)); // fastest way
+		if (is_array($s) || $s instanceof \Countable) {
+			return count($s);
+		} elseif ($s instanceof \Traversable) {
+			return iterator_count($s);
```
hrach: this (may) mutate the iterator, imo it's best to let decide the programmer.

> iterator_count() is not guaranteed to retain the current position of the iterator. 
> http://php.net/iterator_count

PR #105, opened 2016-03-14
added macro {spaceless}

dg:
Removes extra whitespace.

### DIFF
--- a/src/Latte/Macros/CoreMacros.php
+++ b/src/Latte/Macros/CoreMacros.php
@@ -8,6 +8,7 @@
 namespace Latte\Macros;
 
 use Latte;
+use Latte\Engine;
 use Latte\CompileException;
 use Latte\MacroNode;
 use Latte\PhpWriter;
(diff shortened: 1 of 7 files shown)


Majkl578:
:-1: Multiline tags are useful, especially with multiple attributes, incl. many data-*. It makes it more readable too in that case.

dg:
Changed to macro {compress} and n:compress.

src/Latte/Macros/CoreMacros.php
```diff
@@ -26,6 +26,7 @@
  * - {_expression} echo translation with escaping
  * - {attr ?} HTML element attributes
  * - {capture ?} ... {/capture} capture block to parameter
+ * - {compress} ... {/compress} compresses output
```
JanTvrdik: „compresses output“ is confusing, what about „removes insignificant HTML whitespace“

JanTvrdik:
What about more descriptive name `{compressTags}` or `{compressHtmlTags}`. Rarely used macros must have more descriptive names.

dg:
It compressed whitespace ;)

JanTvrdik:
@dg The tests are so confusing that I changed my mind twice about what I think the macro do.

milo:
What about name `shrink`?

dg:
It compresses spaces, so … n:vakuum? :)

Majkl578:
Or `shrinkWhitespaces`?

`compress` is extremely ambiguous, one could think of real compression.

dg:
What about spaceless? https://docs.djangoproject.com/en/dev/ref/templates/builtins/#spaceless

PR #106, opened 2016-03-19
[WIP] Template with {contentType ...} rendered to string should not modify global application state

jkuchar:
Anything rendered to string should not modify global application state. Unfortunately it is not a case with macro `{contentType mime-type}` which sets headers directly. [ref to latte code](https://github.com/nette/latte/blob/62666d98954c92c0a18e55bfa36343dc2a8d8270/src/Latte/Macros/CoreMacros.php#L446)

Expected behaviour would be to:
1. `render()` method modifies application output directly, it is expected to modify output headers
2. `renderToString()` should NOT modify headers - there should be some other API to get headers set by template

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
         - php: hhvm
 
 script:
-    - vendor/bin/tester tests -s -p php -c tests/php-unix.ini
+    - vendor/bin/tester tests -s -p php-cgi -c tests/php-unix.ini
     - php temp/code-checker/src/code-checker.php --short-arrays -i use.latte
 
 after_failure:
--- a/tests/Latte/Engine.renderToString.phpt
+++ b/tests/Latte/Engine.renderToString.phpt
@@ -1,0 +1,24 @@
+<?php
+
+/**
+ * Test: Latte\Engine::renderToString()
+ */
+
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+test(function () {
+	$template_text = __DIR__ . '/templates/contentType-text.latte';
+	$template_mime = __DIR__ . '/templates/contentType-mime.latte';
+
+	$latte = new Latte\Engine;
+	$latte->setTempDirectory(TEMP_DIR);
+
+	Assert::equal("Content\n", $latte->renderToString($template_text));
+	Assert::equal([], headers_list());
+
+	Assert::equal("Content\n", $latte->renderToString($template_mime));
+	Assert::equal([], headers_list()); // should not affect global state
+});
--- a/tests/Latte/templates/contentType-mime.latte
+++ b/tests/Latte/templates/contentType-mime.latte
@@ -1,0 +1,2 @@
+{contentType plain/text}
+Content
--- a/tests/Latte/templates/contentType-text.latte
+++ b/tests/Latte/templates/contentType-text.latte
@@ -1,0 +1,2 @@
+{contentType text}
+Content
--- a/tests/php-unix.ini
+++ b/tests/php-unix.ini
@@ -1,4 +1,6 @@
 [PHP]
+expose_php = off
+
 ;extension_dir = "./ext"
 
 [Zend]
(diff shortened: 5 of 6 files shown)


fprochazka:
Sounds logical. Imho sending of the header should be removed altogether, as "sending a headers from template engine" sounds very wrong when you say it out loud like this. Sadly, that would be a giant BC Break.

jkuchar:
**render()** method is dependent on script output anyway. It is convenient when it sends headers in render() method. You need to send headers before output and it will be really tricky to implement because you know content type after actual template execution.

**renderToString()**: It should just return rendering output.

There can introduced something like `renderToBuffer()` which returns DTO with "response body" and headers.

jkuchar:
Alternate solution can be to declare content type in some other **static** way.

E.g. using template headers (something very similar to file-level docblock)

Then:
- `render()` then will not need to send headers anymore, because it will be easy to determine content-type before template execution
- `renderToString()` will remains the same
- there will be need for API to get content type of given latte file

fprochazka:
In Lavito, we've implemented "the symfony way" of naming templates (while integrating latte into symfony)... that means `.txt.latte` is content type text, `.html.latte` is content type html. Imho it's cleaner, since you set the content type from outside, and you can read it easily from the prepared (and not yet rendered) object and send the header yourself if you must.

We could also make it that you have to set the type from presenter and deprecate the macro - which would solve the chicken egg problem of fetching the content type, sending the header and rendering the template.

I agree the `render()` should not be changed for the compatibility's sake. But still I think in `3.*` releases it might be a good idea to break that and it's good that you brought up this issue.

fprochazka:
> I think it is great that latte file knows it's content type. That's why latte is what it is - content escaping.

I agree, I just think the macro shouldn't be sending HTTP Headers.

---

In Lavito, we've implemented "the symfony way" of naming templates (while integrating latte into symfony)... that means `.txt.latte` is content type text, `.html.latte` is content type html. Imho it's cleaner, since you set the content type from outside, and you can read it easily from the prepared (and not yet rendered) object and send the header yourself if you must. We're using it for email templates, that are not yet in HTML, but plain. And we wouldn't want that to be sending any headers to the output - makes no sense.

> it will be really tricky to implement without sending it directly from template because you know content type only at runtime

I disagree. We could make it that you have to set the type from control/presenter (some template object setter - compiler has that ability already) and deprecate the macro - which would solve the chicken egg problem of fetching the content type, sending the header and rendering the template. The real BC Break would be only those few templates where you have `xml` content-type macro for your exports/rss - and it can be easily fixed, since you can grep your project for the macro and find the template usage.

---

I agree the `render()` should not be changed for the compatibility's sake. But still I think in `3.*` releases it might be a good idea to break that and it's good that you brought up this issue.

jkuchar:
> We could also make it that you have to set the type from control/presenter (some template object setter - compiler has that ability already)

I do not agree with this. I has two downsites:
- no way for static analysis to know what content type is inside
- you can interpret one latte template from to places with two different content types which can allow you do XSS

-----------

Template **for itself** must know what is inside. I think there are two meaningfull ways of doing that:
- filename as proposed by Filip
  - plus: easy to implement + easy to understand
  - minus: file-system dependency ([compiler](https://api.nette.org/2.3.9/source-Latte.Compiler.php.html#88-137) should has content type as parameter and [Engine](https://api.nette.org/2.3.9/Latte.Engine.html] should determine content type from filename)
- by introducing latte header (e.g. must be first macro in file, etc.)
  - plus: latte itself knows what is inside; there will be not API change to `Compiler:compile()` and `Engine`
  - plus: contentType can act as "header macro" (first macro in latte; no variables) will be easy to introduce without BC
  - when using PresenterResponse there will be no BC, because it will send header retrieved from template
  - minus: requires to add new syntax into latte

dg:
Just note: `{contentType}` sends HTTP header only when you pass full mimetype ie `{contentType application/javascript}`. For content sensitive escaping is enough to specify `{contentType javascript}` and no header is sent.

PR #109, opened 2016-04-27
Filters: more efficient filters cache for non-lowercased filters

JanTvrdik:


dg:
This makes slower add() (which is used everytime) and makes non-lowercased filter (which are used sparingly) faster only nonmeasurable.

JanTvrdik:
The `add` is slower only when filters is rewritten which is rare.

dg:
IMHO this is not area where performance can be improved. It is nearly ideal.

I thought about about it caching for dynamic filters:

```php
		return $this->$name = function ($arg) use ($lname, $name, & $cache) {
			$args = func_get_args();
			array_unshift($args, $lname);
			if ($cache) {
				return call_user_func_array($cache, $args);
			}
			foreach ($this->_dynamic as $filter) {
				$res = call_user_func_array(Helpers::checkCallback($filter), $args);
				if ($res !== NULL) {
					$cache = $filter;
					return $res;
				} elseif (isset($this->_static[$lname])) {
					$this->$name = Helpers::checkCallback($this->_static[$lname]);
					return call_user_func_array($this->$name, func_get_args());
				}
			}
			$hint = ($t = Helpers::getSuggestion(array_keys($this->_static), $name)) ? ", did you mean '$t'?" : '.';
			throw new \LogicException("Filter '$name' is not defined$hint");
		};
```

JanTvrdik:
> IMHO this is not area where performance can be improved

That's my specialty, improving performance of stuff that are not important :laughing: 

---

BTW: Why don't we just lower-case filter calls during compilation (emit `$this->filters->escapejs($var)` instead of `$this->filters->escapeJs($var)`)

dg:
Yes, it is possible.

JanTvrdik:
ad dynamic filters

~~~php
return $this->$name = function ($arg) use ($lname, $name) {
	$args = func_get_args();
	array_unshift($args, $lname);
	foreach ($this->_dynamic as $filter) {
		$res = call_user_func_array(Helpers::checkCallback($filter), $args);
		if ($res !== NULL) {


			$this->$name = function ($arg) use ($lname, $filter) {
				$args = func_get_args();
				array_unshift($args, $lname);
				return call_user_func_array($filter, $args);
			};


			return $res;
		} elseif (isset($this->_static[$lname])) {
			Helpers::checkCallback($this->_static[$lname]);
			$this->$name = & $this->_static[$lname];
			return call_user_func_array($this->$name, func_get_args());
		}
	}
	$hint = ($t = Helpers::getSuggestion(array_keys($this->_static), $name)) ? ", did you mean '$t'?" : '.';
	throw new \LogicException("Filter '$name' is not defined$hint");
};
~~~

dg:
I see no difference…

JanTvrdik:
I removed the `if` call =)

dg:
Oh!

dg:
But I am not sure whether it is BC break or not…

JanTvrdik:
Could you elaborate? I see no BC break.

dg:
One can create dynamic filter which accepts or rejects (ie return NULL) accoring its parameters. theoretically

JanTvrdik:
Good point, I guess I would give up on optimizing dynamic filters, people should use static filters or register static filter inside dynamic filter call if they care about performance.

dg:
Yes, thats why I didn't commit it.

dg:
Other posibitlity is to make filters case sensitive and add some of them in multiple variants.

JanTvrdik:
You mean like add both `escapeJs` and `escapeJS`? Yes, that would work. Although I think that with the suggestions in the exception message, allowing only `escapeJs` would cause a lot less paint than before. It would still be a BC break though. I think that lower-casing during compilation is the best BCbreak-free option.

PR #114, opened 2016-05-05
PhpWriter: fixed handling of uppercase properties [closes #113]

### LINKED ISSUE #113: Properties of objects in allcaps are refused during compilation time (throws user deprecated error)
| |  |
|-------------------|--------------|
| **Type**          | Bug |
| **Broken versions**          | 2.4.0-RC1, dev-master |
| **OK versions**          | <= 2.3.11 |

When you have `{$variable->ANYTHING}` in your latte template, you'll get a `E_USER_DEPRECATED` error `Replace literal ANYTHING with constant('ANYTHING')`

This error is thrown at template compile time, by [`Latte\PhpWriter.php:182`](https://github.com/nette/latte/blob/master/src/Latte/PhpWriter.php#L182)

**Test code:**
```
{? $testVar1 = (object)['TESTATTRIBUTE' => 'test value 1']}
{? $testVar2 = (object)['testattribute' => 'test value 2']}
<div n:if="!empty($testVar1)">
    {$testVar1->TESTATTRIBUTE}
</div>
```

**Current result:**
```
User Deprecated

Replace literal TESTATTRIBUTE with constant('TESTATTRIBUTE')
```

In my use-case, the variable is instance of `Dibi\Row` and the all-caps attribute is a column name returned by ODBC driver, which doesn't support `COLUMN as column` syntax :smiley: :cry:

JanTvrdik:


dg:
Thanks!

PR #116, opened 2016-05-13
Content aware filters WIP

dg:
https://forum.nette.org/cs/26286-latte-a-plaintextove-versus-blokove-filtry


dg:
I think that silent conversion html -> text would not be there.

src/Latte/Runtime/FilterInfo.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+class FilterInfo
+{
+	use Latte\Strict;
+
+	/** @var string|NULL */
+	public $contentType;
+
+
+	public function __construct($contentType = NULL)
```
fprochazka: is it neccesary to create new instance every time? Imho this could be made into [flyweight](https://en.wikipedia.org/wiki/Flyweight_pattern).

dg: It might, but creating an instance is extreme cheap.

PR #117, opened 2016-05-16
added invocation resolver [WIP]

dg:


Majkl578:
:-1: for bare callable. Why not an interface?

fprochazka:
Do I understand correctly, that this can inline values of function calls or expressions into the template, so they don't have to be called each time the template is rendered? If so, epic! This is perfect for translation macros (which would still need to distinguish templates based on selected language, but still..) :+1:

dg:
Unfortunately not :-) The original intention was to have the ability to create pseudoglobal functions (eg `xyz()` in `{$a + xyz($b)}`), maybe all filters can be functions too, the second intention was to control which functions/method may be called (sandboxing, whitelist, blacklist), and eventually be able to create [pseudoconstants](https://github.com/nette/nette/pull/1122#issuecomment-21101977), but this is not covered by this first commit.

Support for inlining translation macros for static data is in fact allowed by this commit 9ee17b572a1f617c55f32b6e4f71a8117d7f4957.

xificurk:
> maybe all filters can be functions too

yes, please :heart:

dg:
@xificurk It is not too useful now, because you can use modifiers everywhere `{$a + ($b|xyz)}`

Majkl578:
@dg: So now you are deleting comments you disagree with? Or which disagree with you? Nice censorship you have here.

dg:
Your comment started with :-1: and was negative. Actually I didn't read it.

JanTvrdik:
@dg This would make our coders very happy. Current filters are like methods on objects (you have a base value on which you call some transformation). But sometimes there is no base value, in PHP you would create a static method, in Latte you currently have no good solution. Sometime you can define a block but that does not work well in all cases.

TomasVotruba:
What needs to be done here?

dg:
It is targeted to next version.

PR #120, opened 2016-06-06
Latte\Engine: add getProviders method

mrtnzlml:
This method allows to get all providers passed to the Latte\Engine using method Latte\Engine::addProvider. It useful for example when you have custom Template object and you need to work there with providers.

see: https://forum.nette.org/cs/26250-pojdte-otestovat-nette-2-4-rc?p=3#p175305

### DIFF
--- a/src/Latte/Engine.php
+++ b/src/Latte/Engine.php
@@ -294,6 +294,16 @@ class Engine
 
 
 	/**
+	 * Returns all providers.
+	 * @return array
+	 */
+	public function getProviders()
+	{
+		return $this->providers;
+	}
+
+
+	/**
 	 * @return self
 	 */
 	public function setContentType($type)
--- a/tests/Latte/Engine.providers.phpt
+++ b/tests/Latte/Engine.providers.phpt
@@ -1,0 +1,19 @@
+<?php
+
+use Tester\Assert;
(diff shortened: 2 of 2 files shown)


tests/Latte/Engine.providers.phpt
```diff
@@ -0,0 +1,20 @@
+<?php
+
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+
+$latte = new Latte\Engine;
+$latte->setLoader(new Latte\Loaders\StringLoader);
```
mrtnzlml: Not needed - working on it.

src/Latte/Engine.php
```diff
@@ -294,6 +294,16 @@ public function addProvider($name, $value)
 
 
 	/**
+	 * Returns all providers.
+	 * @return string[]
```
dg: `@return array`

mrtnzlml: Ok I saw this above. What is difference between `string[]` and `array`? `string[]` is only array of strings without keys?

mrtnzlml: fixed

dg:
Thanks

PR #121, opened 2016-06-07
Filters: remove use of length from truncate & substring methods

davidkrmela:
`Filters::truncate` & `Filters::substring` should not use `Filters::length` because posibility of object which is `Countable` or `Traversable` and has magic method `__toString`.

For example, in Latte `{$article}` works well, but `{$article|truncate:20}` don't and that is imho inconsistent behaviour. What do you think?


dg:
Nice catch, thanks.

PR #124, opened 2016-06-28
CoreMacros: translate macro does not use output buffer when possible

JanTvrdik:
I wonder if we can drop FilterInfo as well for static content.


dg:
Thank you

PR #125, opened 2016-07-04
BlockMacros: fixed expandTokens in dynamic blocks

matej21:
https://forum.nette.org/en/26640-n-tags-showing-upin-source-if-block-name-is-a-variable-latte-standalone


dg:
Thanks!

PR #127, opened 2016-07-29
Snippets: disable snippet mode when rendering snippets

matej21:
https://forum.nette.org/cs/26805-sekundarni-sablona-komponenty-ignoruje-makro-extends-pri-ajaxu

But it is a bc break. Is that ok with you?


dg:
IMHO it is ok, thanks

PR #131, opened 2016-08-15
BlockMacros - define: fixed variables extracting

matej21:
...when named parameters are used

currently, even global variables like `$basePath` are not available...


dg:
That was the intention ... I understand that in some situations it may be undesirable.

matej21:
I was afraid of this answer. So what do you suggest? What about extracting only global params?

dg:
Global params seems like good compromise.

matej21:
updated

dg:
Thanks

PR #132, opened 2016-08-21
Auto empty macros: fix combination with a short macro closing

matej21:
e.g. `{label foo}bar{/}`

there is a minor BC break when someone adds macro directly using a addMacro (without MacroSet) without passing correct PAIR flag.

I really don't like hardcoded `_` macro. Maybe we could introduce another flag for this?


JanTvrdik:
> Maybe we could introduce another flag for this?

Because we have too little flags.

dg:
Dynamic nature of macros is problematic and something like `PAIR` is needed. For consistency better name is `EMPTY` (for opposite), but constant cannot be `EMPTY` in PHP < 7. 

`$node->empty` should be set in `Compiler::openMacro` according to this flag.

dg:
Auto-empty macros are deprecated and removed in Latte 3

PR #139, opened 2016-10-01
Make normalizePath() not forget leading ../

### LINKED ISSUE #138: FileLoader::normalizePath() strips leading ../
First, thanks for making Latte! ☕ 

While migrating to Latte 2.4 I've find the following issue, if I do this in a presenter:

```
$this->createTemplate()->setFile('../../foo/bar.latte);
```

and there's
```
{include waldo.latte}
```

in `bar.latte`, rendering fails with

***RuntimeException: Missing template file foo/waldo.latte***

I've tracked it down to an issue in [`Latte\Loaders\FileLoader::normalizePath()`](https://github.com/nette/latte/blob/003e4849365e28d4a78bef42c2c2e071d647c526/src/Latte/Loaders/FileLoader.php#L87-L98) where in the first [`foreach`](https://github.com/nette/latte/blob/003e4849365e28d4a78bef42c2c2e071d647c526/src/Latte/Loaders/FileLoader.php#L90) iteration when `$res` is still empty then if it hits [`$part === '..'`](https://github.com/nette/latte/blob/003e4849365e28d4a78bef42c2c2e071d647c526/src/Latte/Loaders/FileLoader.php#L91) condition then the `$part` is just thrown away because it tries to [`array_pop`](https://github.com/nette/latte/blob/003e4849365e28d4a78bef42c2c2e071d647c526/src/Latte/Loaders/FileLoader.php#L92) from an empty array.

The issue can be demonstrated by adding
```
Assert::same('../../tests', $loader->getReferredName('../tests', '../file'));
```
to [Loaders.FileLoader.phpt](https://github.com/nette/latte/blob/003e4849365e28d4a78bef42c2c2e071d647c526/tests/Latte/Loaders.FileLoader.phpt
).

Hope this is at least a bit understandable.

The workaround is to never pass leading `../` to `setFile()` (and use `realpath()` to remove these) but since I've already spent quite a while figuring out where's the problem, the proposed fix is an incoming pull request.

spaze:
Fixes #138 by just remembering these `..` parts and the add them to the rest when normalizing the path

Thanks for considering this fix!

### DIFF
--- a/src/Latte/Loaders/FileLoader.php
+++ b/src/Latte/Loaders/FileLoader.php
@@ -86,15 +86,20 @@ class FileLoader implements Latte\ILoader
 	 */
 	private static function normalizePath($path)
 	{
-		$res = [];
+		$res = $prefixes = [];
 		foreach (explode('/', strtr($path, '\\', '/')) as $part) {
-			if ($part === '..' && $res) {
-				array_pop($res);
+			if ($part === '..') {
+				if ($res) {
+					array_pop($res);
+				} else {
+					$prefixes[] = $part;
+				}
 			} elseif ($part !== '.') {
 				$res[] = $part;
 			}
 		}
-		return implode(DIRECTORY_SEPARATOR, $res);
+		return implode(DIRECTORY_SEPARATOR, array_merge($prefixes, $res));
+
 	}
 
 }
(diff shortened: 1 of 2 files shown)


spaze:
@dg That's much nicer, indeed! Thanks!

PR #144, opened 2016-10-29
Case-sensitive filters

EdaCZ:
Hi. I tryed to prepare pull request with case-sensitive filters. I hope it is ok.

Motivation:
1. Almost every part of Nette is case sensitive now, so filters should be sensitive too.
2. Its cleaner when exists the only one way to call filter (now you can call `{$var|leftPad}` and also `{$var|leftpad}`, `{$var|Leftpad}` etc.).
3. Now, when you register user static filter (for example `myFilter`) and call it in Latte with proper camel case (exactly as you register it, for example `{$var|myFilter}`), in every single filter call magic method FilterExecutor::__get() is called, although when calling lowercase its is called directly via property (except the first call). From my point of view this is not correct that calling filter in lowercase ("bad cases") is quicker than "proper case" call.
4. Code is now a little bit nicer.
5. It opens doors to next step (and simplify its implementation) - full support of dynamic macros in method FilterExecutor::renderContent() I want to implement.

Notes:
- Back compatibility of builtin Nette filters is implemented via property $_deprecated with deprecation notice.
- If its not ok to deprecate lowercase filters, i can move lowercase filters back to array _static so we will support it too.
- When you use user-defined macro, it throws Exception with correct notice, that you are calling filter with "bad cases".
- What with `substr` filter? I think good camel case name should be `subString` (see same name method in class `Filters`), but someone maybe want to use it with same name as PHP function substr. What do you think?
- Yes, it is BC break, but I think it is worth it. Same as case sensitivity in Routing, which is allready implemented.

If you have any suggestions, don't hesitate to comment :-)


TomasVotruba:
Hi, the commit is too long and difficult to read.
Could you split it to more commits?

TomasVotruba:
Could you also reference the tweet that started this? Add the picture to the issue if you can.

EdaCZ:
Ok, I am trying (so hard :-) - tests must passed after each commit) to split it now.

Tweets are here:
https://twitter.com/EdaCZ/status/791229264578445312
https://twitter.com/EdaCZ/status/791234975408480256

TomasVotruba:
@EdaCZ Thanks. I know it can be messy sometimes :)

> But tweets and image is the next step. It should be other issue, after this will be resolved/rejected.

Maybe create an issue and reference this PR there then. So it's clear to all who didn't see that on Twitter.

EdaCZ:
So, here we are. Is it better? If you want me to explain something it is not problem.

src/Latte/Runtime/FilterExecutor.php
```diff
@@ -113,24 +109,28 @@ public function __get($name)
 						: $res;
 				};
 			} else { // classic filter
-				return $this->$lname = $callback;
+				return $this->$name = $callback;
 			}
 		}
 
-		return $this->$lname = function ($arg) use ($lname, $name) { // dynamic filter
+		return $this->$name = function ($arg) use ($name, $name) { // dynamic filter
```
TomasVotruba: one `$name` can be removed

TomasVotruba:
Much better, thank you. Now I understand even without reading a PR description :) Awesome job!

@dg Could you review?

dg:
I am still not sure whether case sensitive filters are really good idea.

f3l1x:
I vote for case insensitive.

EdaCZ:
And where is the difference between case sensitive filters and case sensitive routes?

Almost every part of Nette is case sensitive, filters not.

TomasVotruba:
I'm for case sensitivity, since ambiguity will eventually lead to problems.

dg:
I think that this implementation doesn't provide good back compatibility (will |anYCase work for user defined filter |anyCase?) and should trigger better error messages (should tell exactly what to use).

For some commonly used modifiers like |stripTags, |noCheck etc. should IMHO work lowercased variant too, without error.

EdaCZ:
Ok, fine. I will improve back compatibility.

dg:
@EdaCZ do you have any progress?

dg:
Ping @EdaCZ

EdaCZ:
@dg I'll try to look on it at the weekend.

TheDigitalOrchard:
I don't agree with case-sensitive filters. Latte is designed to mirror PHP closely, and neither functions nor class methods are case-sensitive. Filters should not be.

dg:
Implemented in Latte 3

Issue #149, opened 2017-01-05
Incorrect enclosing of string inside javascript strings


josefsabl:
- bug report? yes (at least I think so :)
- feature request? no
- version: 2.4.2

### Description
String variables are enclosed by `&quot;` even if inside quotes already (string) in Javascript generating duplicate quotes resulting in unwanted quotes inside string (in better case) or even broken Javascript.

### Steps To Reproduce
```
{var $world = "World"}

<button onclick="alert('Hello {$world}!')">HW</button>
<button onclick='alert("Hello {$world}!")'>HW</button>
```

generates

```
<button onclick="alert('hello &quot;World&quot;')">HW</button> displays Hello "World"! (notice the unwanted quotes)
<button onclick='alert("hello &quot;World&quot;")'>HW</button> generates syntax error
```

JanTvrdik:
The code you have written is incorrect, it should be 

~~~html
{var $world = "World"}

<button onclick="alert('Hello ' + {$world} + '!')">HW</button>
<button onclick='alert("Hello " + {$world} + "!")'>HW</button>
~~~

josefsabl:
Thanks for the response. So it is not a bug but missing feature?

It definitely does not respect this claim from [docs](https://latte.nette.org/en/#toc-context-aware-escaping): **All is done automatically, consistently and correctly.**

The resulting HTML is not okay either: `<button onclick="alert('Hello ' + $quot;World$quot; + '!')">HW</button>` vs `<button onclick="alert('Hello World!')">HW</button>`

And also note that latte is templating engine and imho shouldn't require changes to the surrounding code. From this perspective this behavior is non-standard at least. In other templating language (PHP e.g. ;) you can definitely write:

`<button onclick="alert('Hello <?= htmlspecialchars($world) ?>!')">HW</button>`

It will work just fine and you wouldn't need to change JS code itself.

Just asking, no offence intended :-)
<3

dg:
> you can definitely write:
> <button onclick="alert('Hello <?= htmlspecialchars($world) ?>!')">HW</button>

You can, but it is XSS. `$world` can contain `'); destroyComputer('`.

It's true that this is missing in documentation, could you please add it?

josefsabl:
Ehh, sure, that is why I use latte :1st_place_medal: 

doc: can do

josefsabl:
Anyways, the fact that there is XSS in my example doesn't render the rest of comment invalid, does it? :-)

dg:
[...]

PR #150, opened 2017-01-05
Added limitation of escaping in JavaScript to readme.md

josefsabl:
- bug fix? https://github.com/nette/latte/issues/149   
- new feature? no
- BC break? no
- doc PR: latte/docs/#toc-context-aware-escaping???  

As discussed here: https://github.com/nette/latte/issues/149


dg:
Thanks!

PR #152, opened 2017-01-17
added {include with blocks} as a successor to {includeblock}

dg:
- bug fix? ~
- new feature? yes
- BC break? no

There are use cases where `{includeblock}` is useful and cannot be substituted by `{import}`. But the name `{includeblock}` is misleading. This is attempt to solve it by more understandable `{include with blocks 'file.latte'}`

Cons: this introduces syntax which is not common in Latte.

https://forum.nette.org/cs/26699-latte-vypis-bloku-do-include

### DIFF
--- a/src/Latte/Macros/BlockMacros.php
+++ b/src/Latte/Macros/BlockMacros.php
@@ -186,7 +186,7 @@ class BlockMacros extends MacroSet
 	 */
 	public function macroIncludeBlock(MacroNode $node, PhpWriter $writer): string
 	{
-		//trigger_error('Tag {includeblock} is deprecated, use similar tag {import}.', E_USER_DEPRECATED);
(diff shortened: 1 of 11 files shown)


JanTvrdik:
I hate the syntax. Also include related stuff are changed way too often -> deeper analysis is required.

dg:
Do you have a better idea on syntax?

JanTvrdik:
Not yet. I don't want to rush this change just to change it in two weeks again.

dg:
I don't know what are you talking about.

f3l1x:
The syntax `{include-with-blocks}` is forbidden?

dg:
`{include-with-blocks}` is not common in Latte too. More expectable is `{includewithblocks}` (hard to read) or `{includeWithBlocks}`.

dg:
Changed to `{include file.latte with blocks}`

PR #154, opened 2017-03-03
Notice fix on PHP7

AdamChysky:
Absence of the type cast caused Notice (A non well formed numeric value encountered) to appear in the log file on PHP7 server.

- bug fix? yes
- new feature? no
- BC break? no


dg:
Nette 2.2 is not supported anymore since may 2016, sorry.

PR #156, opened 2017-04-04
n:class supports BEM

kukulich:
- bug fix? yes
- new feature? no
- BC break? no

[BEM](http://getbem.com/naming/) use `--`  for modifiers. Now it's possible to use classes like `block--hidden` in `n:class` without quotes.


dg:
Thanks

PR #157, opened 2017-04-06
feature: added n:attr-expand macro

dakujem:
The macro expands an array of attributes, similarly to n:attr.

- bug fix? no   
- new feature? yes
- BC break? no
- doc PR: nette/docs#???  

**Motivation**:
Suppose there is an array `$attributes` containing attributes for `<meta>` tag.
```php
array (
  'content' => 'width=device-width, initial-scale=1',
  'name' => 'viewport',
  'http-equiv' => NULL,
  'scheme' => NULL,
)
```		
The developer wants to achieve this HTML output:
```html
<meta content="width=device-width, initial-scale=1" name="viewport">
```
However, using `<meta n:attr="$meta">` the actual output is:
```html
<meta 0="content:width=device-width, initial-scale=1 name:viewport">
```
The developer is currently forced to use workarounds, like:
```latte
{=Nette\Utils\Html::el('meta', $attributes)}
```

**The solution**:
If attributes for a tag are present in an array, it would be simple to create the tag with an array-expanding macro:
```latte
<meta n:attr-expand="$attributes">
```
The macro itself is trivial:
```php
'echo LR\Filters::htmlAttributes( %node.word );'
```
compred to `n:attr`:
```php
'echo LR\Filters::htmlAttributes( %node.array );'
```

**Sidenote**:
In this case, the macro is used for outputting metas:
```latte
<meta n:foreach="$metas as $meta" n:attr-expand="$meta">
```
This approach is however very useful for many other cases during output generation.

### DIFF
--- a/src/Latte/Macros/CoreMacros.php
+++ b/src/Latte/Macros/CoreMacros.php
@@ -84,6 +84,7 @@ class CoreMacros extends MacroSet
 
 		$me->addMacro('class', NULL, NULL, [$me, 'macroClass']);
 		$me->addMacro('attr', NULL, NULL, [$me, 'macroAttr']);
+		$me->addMacro('attr-expand', NULL, NULL, [$me, 'macroAttrExpand']);
 	}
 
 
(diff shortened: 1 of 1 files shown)


matej21:
you can use `n:attr="(expand) $attributes"` :)

dakujem:
Can you please point me towards an entry in documentation about this? I seem not to be able to find it.

dakujem:
Okay, so I finally found the reference myself (by pulling `nette/docs` repo and searching in source)...
The `expand` operator is mentioned in a completely unrelated section of quite unrelated chapter about MVC and Presenters generally: https://doc.nette.org/en/2.4/presenters#toc-links-in-templates

I do believe that this operator should be mentioned in the general Latte syntax description and have it's own description with the Latte macros, because that is where people search for macros and other features of Latte.

matej21:
@dakujem you are right. will you please send a PR to the doc?

TomasVotruba:
@matej21 Where exactly do you think this should be put in? Link to the file would be best.

dg:
Closed by #158

TomasVotruba:
Awesome, thanks @dg

PR #159, opened 2017-04-15
Filters::stripTags() allowed tags parameter support

jaroslavlibal:
The second parameter (`$allowable_tags`) for the `strip_tags` php function recall is not supported since a4bb7fe209a843d94bb6ddb60c995f11c819bdd2.

The support is not dropped from PHP so it should be supported in the filter.

### DIFF
--- a/src/Latte/Runtime/Filters.php
+++ b/src/Latte/Runtime/Filters.php
@@ -212,14 +212,15 @@ class Filters
 	 * Removes tags from HTML (but remains HTML entites).
 	 * @param
 	 * @param  string HTML
+	 * @param  string tags which should not be stripped
 	 * @return string HTML
 	 */
-	public static function stripTags(FilterInfo $info, $s): string
+	public static function stripTags(FilterInfo $info, $s, $allowableTags = NULL): string
 	{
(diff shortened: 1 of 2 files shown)


dg:
Thats right that dropped support for allowable_tags is unexpected BC break. However, I think that filtering of certain tags is not a task for templating system. What is your use case?

jaroslavlibal:
@dg My usecase is to simply allow only the `<strong>` and `<p>` elements (and strip the rest) in the text announcements component (messages are editable by users) for some views.

Not sure, but you can probably say that it is not a task for the templating system for most of the "modification" filters.

However my primary motivation for this PR was that I noticed that it does not work since the mentioned commit and I thought it has been forgotten by a mistake. So I tried to prepare the fix instead of reporting an issue.

Never mind if this PR will be closed without merge, I'm happy for your previous comment as I can't take any lessons just from the thumb down votes :-)

dg:
Problem is that striptags is really stupid function. Input like `<p><ul></ul>Text` is the same as `<p></p><ul></ul>Text`, so the first one will be filtered to `<p>Text`, but the correct is `<p></p>Text`.

I don't want natively support something that is broken, so I'd rather not merge it.

Workaround is simple: `{strip_tags($var, '<p><strong>')|noescape} or you can register own filter striptags and replace the default one.

jaroslavlibal:
Yes, I already did that. Let's close it then, thanks!

PR #161, opened 2017-05-21
Engine::compile() fixed construction of CompileException, doesn't acc…

MartkCz:
…ept nullable type

- bug fix? yes
- new feature? no
- BC break? no


dg:
Thanks

PR #163, opened 2017-06-14
[PROOF OF CONCEPT] change variable passing logic

matej21:
- BC break? yes

Latte is a bit inconsistent how it passes variables to blocks (get_defined_vars vs this->params). Mostly it is "just" a confusing feature, but sometimes it is a bug.

This PR unifies the behavior that only variables from this->params are implicitly available. 

It is done by passing nothing to renderBlock and instead `extract($this->params)` is called in each block. This approach allows future feature to have different parameters for main template and layout. (because it extracts parameters of own Template object)

Of course, there are some BC breaks - mainly that you have to explicitly name parameters you want in a block from the outer scope.

So instead of
```latte
{var $foo = 1}
{block bar}
{$foo}
{/block}
```
you have to write
```latte
{var $foo = 1}
{block bar $foo}
{$foo}
{/block}
```

what do you think about it? 

cc @JanTvrdik

### DIFF
--- a/src/Latte/Macros/BlockMacros.php
+++ b/src/Latte/Macros/BlockMacros.php
@@ -287,7 +287,7 @@ class BlockMacros extends MacroSet
 		}
 		$this->blockTypes[$name] = implode($node->context);
 
-		$include = '$this->renderBlock(%var, ' . (($node->name === 'snippet' || $node->name === 'snippetArea') ? '$this->params' : 'get_defined_vars()')
+		$include = '$this->renderBlock(%var, ' . (($node->name === 'snippet' || $node->name === 'snippetArea') ? '[]' : 'get_defined_vars()')
 			. ($node->modifiers ? ', function ($s, $type) { $_fi = new LR\FilterInfo($type); return %modifyContent($s); }' : '') . ')';
 
 		if ($node->name === 'snippet') {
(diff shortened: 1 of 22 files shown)


f3l1x:
Wow. It's solid BC break. :-) 

What is your motivation to do it? I mean, is it not a expected behaviour?

jkuchar:
It makes sense to me. I would like to explicitly pass variables. Maybe also variable types in future?

dakur:
I'm in favor of whatever explicit control of passing variables. Magic behavior is somewhat confusing to me, code is worse analyzable (e.g. passing variables to includes) and may lead to bugs.

dg:
There is a common practice in web development to mark a part of a template as a block so that you can either overwrite or reuse it. If at that moment the variables in such a delimited part stopped working and one would have to list them similarly like in clause `use` in PHP, it would only be annoying. In fact, even PHP itself moves back from clause `use` thanks to arrow functions.

I don't think there's anything wrong with automatically passing variables to embedded blocks.

Otherwise, even such a basic thing as this will stop working: `<h1 n:block="title">Hello {$world}</h1>`.

What could change is that the macro {include block} would not send all the variables, even if the block is defined in the same file. Except for this `{include parent}`, because it is necessary for overridings.

PR #170, opened 2017-09-18
Show template path

nechutny:
- bug fix? no
- new feature? yes
- BC break? yes 

This commit add HTML comment to rendered template with path of template file.
It's very usefull when developer is fixing bugs in someone else code.
Current implementation is naive and easy - use Tracy Debugger's method isEnabled for enabling.
It also doesn't have context, so it can potentially break included JS/CSS. We can discuss better and more clean solution.

### DIFF
--- a/src/Latte/Loaders/FileLoader.php
+++ b/src/Latte/Loaders/FileLoader.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace Latte\Loaders;
 
 use Latte;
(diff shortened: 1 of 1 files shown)


CZechBoY:
Why it is enabled only for debug? I think you can enable it in production too (and drop dependency to Tracy).

f3l1x:
Well, I think it is great idea but not everyone want it **on**.

What about some `DebugFileLoader`?

PR #171, opened 2017-11-28
[2.4] Hotfix - `isExpired` correct file path

webimpress:
Method was called from `getContent` with the full file path (see lines 36 and 43) and inside the method `isExpired` baseDir was added another time (see line 57), so as the result we have: "baseDir/baseDir/fileName.ext"

https://github.com/nette/latte/blob/e6ab6ccc5c372036b99bf8b03c2b6ff9f1cec621/src/Latte/Loaders/FileLoader.php#L34-L43

https://github.com/nette/latte/blob/e6ab6ccc5c372036b99bf8b03c2b6ff9f1cec621/src/Latte/Loaders/FileLoader.php#L55-L57

Fix for *2.4* branch, but it should be also fixed on master.

- bug fix? yes/~no~   
- new feature? ~yes~/no
- BC break? ~yes~/no
- doc PR: ~nette/docs#???~

### DIFF
--- a/src/Latte/Loaders/FileLoader.php
+++ b/src/Latte/Loaders/FileLoader.php
@@ -31,16 +31,16 @@ class FileLoader implements Latte\ILoader
 	 * Returns template source code.
 	 * @return string
 	 */
-	public function getContent($file)
+	public function getContent($fileName)
 	{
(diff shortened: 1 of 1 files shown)


dg:
Thanks!

PR #175, opened 2018-02-08
added CachingIterator::__toString()

vlastavesely:
When an iterator has been used in a template just as `{$iterator}` and not
as `{$iterator->counter}` (as documented), it has thrown a kind of confusing
error:

  Fatal error: Method Latte\Runtime\CachingIterator::__toString() must not
  throw an exception, caught BadMethodCallException: Latte\Runtime\CachingIterator
  does not fetch string value (see CachingIterator::__construct) in *PATH* on line 0

This patch allows conversion to a string by returning the current counter.

- bug fix? yes
- BC break? no


dg:
Thanks

PR #181, opened 2018-06-28
Filters: added decodeentities filter

dakur:
- bug fix / **new feature**?   
- BC break? yes/**no**
- doc PR: nette/docs#???    
Will add when if this feature will be accepted.

In `Filters.php`, there is `stripHtml` filter which uses `html_entity_decode()` and `strip_tags()` PHP functions. There is `striptags` filter as well, which uses only `strip_tags()` PHP function, but no `decodeentities` which would use only `html_entity_decode()` PHP function.

Quite often, I have `&nbsp;` or `&ndash;` HTML entity stored in database and I need to render it on the page correctly. That's exactly where `decodeentities` would be helpful.


dg:
I think having a entities in DB is more like a mistake, so this is not a use case I would like to solve via native Latte filter.

dakur:
We have some apps with semi-automatic content (items are created manually through migrations, but then loaded from DB and used dynamically), I want to have these in correct typographical form. Of course I can copy&paste a character instead of writing an entity there, but how then I differentiate normal space from non-breaking space?

dakur:
I'm not saying it's not better, I'm just saying that when I work with DB manually, "nbsp" is not differentiable from normal space with eye, so I'd like to store "nbsp" as entity in DB to be able to read/edit it. But it's true that as it's not best practice overall, it would be better not to have it as a common stuff in Latte.

PR #182, opened 2018-06-28
Tests: added missing striptags test

dakur:
Just adding missing test case for `striptags` filter.

### DIFF
--- a/tests/Latte/Filters.stripTags().phpt
+++ b/tests/Latte/Filters.stripTags().phpt
@@ -1,0 +1,51 @@
+<?php
+
+/**
+ * Test: Latte\Runtime\Filters::stripTags()
+ */
+
+declare(strict_types=1);
+
+use Latte\Engine;
+use Latte\Runtime\FilterInfo;
+use Latte\Runtime\Filters;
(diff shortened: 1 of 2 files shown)


tests/Latte/Filters.stripTags().phpt
```diff
@@ -0,0 +1,51 @@
...
+
+	Assert::same('', @Filters::stripTags($info, ''));
+	Assert::same('abc', @Filters::stripTags($info, 'abc'));
+	Assert::same('&lt;  c', @Filters::stripTags($info, '&lt; <b> c'));
+});
+
+
+test(function () {
+	$info = new FilterInfo(Engine::CONTENT_HTML);
+	Assert::same('', @Filters::stripTags($info, ''));
```
dg: There shouldn't be @, right?

dg:
Great, thanks!

tests/Latte/Filters.stripTags().phpt
dakur: I've copied it from `stripHtml` test. Fixed.

PR #185, opened 2018-10-29
Allow custom Template implementation

foxycode:
- new feature
- BC break? no

I'd like to use Latte as base for custom templating engine, but I was unable to move on without copying and editing classes from Latte. For example:
* Can't add rewritten `CachingIterator` to `Template` use statements, because `use` macro was removed and `Compiler` adds use statements as text.

### DIFF
--- a/src/Latte/Compiler/Compiler.php
+++ b/src/Latte/Compiler/Compiler.php
@@ -10,6 +10,8 @@ declare(strict_types=1);
 namespace Latte;
 
 
+use Latte\Runtime\ITemplate;
+
 /**
  * Latte compiler.
  */
@@ -74,6 +76,9 @@ class Compiler
 	/** @var string */
 	private $contentType = self::CONTENT_HTML;
 
+	/** @var string */
+	private $templateCLass = Runtime\Template::class;
+
 	/** @var string|null */
 	private $context;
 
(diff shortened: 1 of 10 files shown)


foxycode:
@dg I removed `IEngine` part, so PR is smaller, hope you liked this feature as is. Next PR's will follow.

src/Latte/Compiler/Compiler.php
```diff
@@ -180,9 +206,11 @@ public function compile(array $tokens, string $className): string
 			$members[] = "\n\tfunction $name($method[arguments])\n\t{\n" . ($method['body'] ? "\t\t$method[body]\n" : '') . "\t}";
 		}
 
+		$useNamespaces = constant("{$this->templateCLass}::USE_NAMESPACES");
+
 		return "<?php\n"
-			. "use Latte\\Runtime as LR;\n\n"
-			. "class $className extends Latte\\Runtime\\Template\n{\n"
+			. ($useNamespaces ? Helpers::composeUseNamespaces($useNamespaces) . "\n" : '')
```
k00ni, line 212: `PHP_EOL` instead of `"\n"` could be more robust here.

JanTvrdik: PHP_EOL is strictly worse; you cannot have a random line with inconsistent ending

dg:
I'm closing it because it's very old ;-)

PR #187, opened 2019-01-02
improved file existence checking; use braced vars in strings

TheDigitalOrchard:
Improved file existence checking; use braced vars in strings for improved readability.

- bug fix / new feature?   
- BC break? no
- doc PR: nette/docs#???

### DIFF
--- a/src/Latte/Engine.php
+++ b/src/Latte/Engine.php
@@ -215,7 +215,7 @@ class Engine
 			? preg_replace('#[^\w@.-]+#', '-', substr($m[0], 1)) . '--'
 			: '';
 		$file = "{$this->tempDirectory}/{$base}{$hash}.php";
-		return file_exists($file) ? $file : false;
+		return file_exists($file) ? $file : '';
 	}
 
 
--- a/src/Latte/Engine.php
+++ b/src/Latte/Engine.php
@@ -127,7 +127,7 @@ class Engine
 		}
 
 		if (!preg_match('#\n|\?#', $name)) {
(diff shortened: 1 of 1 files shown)


src/Latte/Engine.php
```diff
@@ -215,7 +215,7 @@ public function getCacheFile(string $name): string
 			? preg_replace('#[^\w@.-]+#', '-', substr($m[0], 1)) . '--'
 			: '';
 		$file = "{$this->tempDirectory}/{$base}{$hash}.php";
-		return file_exists($file) ? $file : false;
```
milo: IMHO,  `null` and `?string` return type is better.

TheDigitalOrchard: What are your thoughts on checking for the existence of the file before trying to read the file?

I'm surprised by the use of the silence @ operator throughout, and in my enviro (and likely some others), it still generates errors due to how the framework in use handles such errors, even when silenced.

milo: Sorry, I overlooked that you added `file_exists()` code. That's wrong. Method only builds the cache file name.

dg:
Whats wrong with current solution? It is really elaborated peace of code.

milo:
Closed for reasons written above.

You can't prevent race conditions without exclusive locking, which is quite expensive for some cases, so usage of `@` operator is appropriate. For example, code proposed by you:
```php
file_exists($file) ? (int)filemtime($file) : 0
```
may throw warning, when `$file` is deleted between `file_exist()` and `filemtime()` calls.

PR #189, opened 2019-02-20
 Filters: remove string type hints from replace filter

pavolbiely:
A long time ago it was possible to use the replace filter like this:
{$var|replace:[' ','+']:['','00']}

Today it ends up like this:
Argument 3 passed to Latte\Runtime\Filters::replace() must be of the type string, array given

Is it possible to avoid this BC break?


dg:
Thanks

CZechBoY:
Isnt `string[]` better than general `array`?

PR #196, opened 2019-08-31
Filters::spacelessHtml() Add support for use <script> as link to JS

janbarasek:
- bug fix
- BC break? yes

In case of `block y` generate too long code:

```html
{block x|strip}
	Lorem ipsum dolor sit amet,
	<script src="main.js"></script>
	consectetur adipiscing elit.
{/block}

------

{block y|strip}
	Lorem ipsum dolor sit amet,
	<script src="main.js"
			title="value"
	></script>
	consectetur adipiscing elit.
{/block}
```

Old return:

```html
Lorem ipsum dolor sit amet, <script src="main.js"></script> consectetur adipiscing elit.
------

Lorem ipsum dolor sit amet, <script src="main.js"
			title="value"
	></script> consectetur adipiscing elit.
```

New return:

```html
Lorem ipsum dolor sit amet, <script src="main.js"></script> consectetur adipiscing elit.
------

Lorem ipsum dolor sit amet, <script src="main.js" title="value"></script> consectetur adipiscing elit.
```

I think this feature is absolutely safe, because if you have empty `<script>` body you can safely replace repeating whitespaces to single whitespace.

Real case in real website:

Snímek obrazovky 2019-08-31 v 17 56 23

Thanks.

### DIFF
--- a/src/Latte/Runtime/Filters.php
+++ b/src/Latte/Runtime/Filters.php
@@ -316,7 +316,7 @@ class Filters
 	 * @param  bool  $strip  stripping mode
 	 * @return string HTML
 	 */
-	public static function spacelessHtml(string $s, int $phase = null, bool &$strip = true): string
+	public static function spacelessHtml(string $s, ?int $phase = null, bool &$strip = true): string
 	{
 		if ($phase & PHP_OUTPUT_HANDLER_START) {
 			$s = ltrim($s);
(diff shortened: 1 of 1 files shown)


src/Latte/Runtime/Filters.php
```diff
@@ -316,26 +316,33 @@ public static function strip(FilterInfo $info, string $s): string
...
 			$s = ltrim($s);
 		}
 		if ($phase & PHP_OUTPUT_HANDLER_FINAL) {
 			$s = rtrim($s);
 		}
-		return preg_replace_callback(
+		$return = (string) preg_replace_callback( // Other cases
 			'#[ \t\r\n]+|<(/)?(textarea|pre|script)(?=\W)#si',
-			function ($m) use (&$strip) {
+			static function (array $m) use (&$strip): string {
```
dg: Why `static`?

src/Latte/Runtime/Filters.php
```diff
@@ -316,26 +316,33 @@ public static function strip(FilterInfo $info, string $s): string
...
-					return $m[0];
+					return (string) $strip ? ' ' : $m[0];
 				}
+				$strip = !empty($m[1]);
+				return (string) $m[0];
 			},
 			$s
 		);
+		$return = (string) preg_replace_callback( // <script> for include JS file
+			'/<script\s*([^>]+?)>(?:\s*)<\/script>/',
```
dg: Will this regexp work with input like `<script title="> <\/script>">` ?

janbarasek: This input is not supported now. I'll try to come up with a better solution.

src/Latte/Runtime/Filters.php
janbarasek: PHP does not have to create a new instance every time.

src/Latte/Runtime/Filters.php
```diff
@@ -316,26 +316,33 @@ public static function strip(FilterInfo $info, string $s): string
...
+	public static function spacelessHtml(string $s, ?int $phase = null, bool &$strip = true): string
 	{
 		if ($phase & PHP_OUTPUT_HANDLER_START) {
 			$s = ltrim($s);
 		}
 		if ($phase & PHP_OUTPUT_HANDLER_FINAL) {
 			$s = rtrim($s);
 		}
-		return preg_replace_callback(
+		$return = (string) preg_replace_callback( // Other cases
```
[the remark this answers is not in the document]
janbarasek: Because PhpStorm hint to me new type can be returned in future. For example `null` in case of error.

PR #197, opened 2019-09-13
Fixed  override in include macro

Jax-p:
- fix, [Nette forum thread](https://forum.nette.org/cs/32540-parametry-pri-pouziti-include-jsou-vzdy-null#p206103)

Instead of overriding `$_args` in the due of `extract($_args)` while having `$_args['$_args']` it extends them.


dg:
Fixed by 276c39052129ef32a084d55156b4d3db2e377466

PR #199, opened 2019-09-24
Fix compatibility with PHP 7.4

Rixafy:
- bug fix
- BC break? no (function is available since 7.1.0)

Function ReflectionType::__toString() is deprecated, use getName() instead.

### DIFF
--- a/src/Latte/Runtime/FilterExecutor.php
+++ b/src/Latte/Runtime/FilterExecutor.php
@@ -178,7 +178,7 @@ class FilterExecutor
 				? new \ReflectionMethod($callback[0], $callback[1])
 				: new \ReflectionFunction($callback);
 			$this->_static[$name][1] = ($tmp = $ref->getParameters())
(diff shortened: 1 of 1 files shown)


src/Latte/Runtime/FilterExecutor.php
```diff
@@ -178,6 +178,7 @@ private function prepareFilter(string $name): array
 				? new \ReflectionMethod($callback[0], $callback[1])
 				: new \ReflectionFunction($callback);
 			$this->_static[$name][1] = ($tmp = $ref->getParameters())
+				&& $tmp[0]->getType() !== null
```
mabar: Btw, did you find why is it null at PHP 7.4?

Rixafy: I think that all tests failed, before getName function, it was casted to string, so it could not be null in if check

dg:
Thanks

Rixafy:
Thanks for merging so quick

PR #207, opened 2020-01-04
Typehint macros

mabar:
- new feature
- BC break? no
- doc PR: let's talk about implementation first

Added support for `{varType string $foo}` and `{templateType TemplateClass}`

TODOs
- `@property` and `@property-read` support for `{templateType}`

Refs
- [typed templates (czech only)](https://phpfashion.com/phpstorm-a-napovidani-nad-this-template)
- [phpstorm latte plugin implementation](https://github.com/mesour/intellij-latte/tree/mn-php-content-and-classes)

cc @mesour

### DIFF
--- a/tests/Latte/CoreMacros.templateType.php74.phpt
+++ b/tests/Latte/CoreMacros.templateType.php74.phpt
@@ -1,0 +1,36 @@
+<?php
+
+/**
+ * Test: Latte\CoreMacros: {templateType ClassWithTypes}
+ */
+
+declare(strict_types=1);
+
+use Latte\Macros\CoreMacros;
+use Tester\Assert;
+use Tester\Environment;
+use Tester\FileMock;
+
+require __DIR__ . '/../bootstrap.php';
+
+if (PHP_VERSION_ID < 70400) {
+	Environment::skip('Typed properties cannot be tested with PHP < 7.4');
+}
+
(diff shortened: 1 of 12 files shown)


tests/Latte/CoreMacros.templateType.php74.phpt
```diff
@@ -0,0 +1,36 @@
...
+if (PHP_VERSION_ID < 70400) {
+	Environment::skip('Typed properties cannot be tested with PHP < 7.4');
+}
+
+$compiler = new Latte\Compiler;
+CoreMacros::install($compiler);
+
+test(function () use ($compiler) {
+	require FileMock::create('
+	class TemplateTypeClass
```
mabar: Not sure if it's possible to define class conditionally. Test fails for php < 7.4 before Environment::skip() is called

dg:
I have added macros `{varType}` and `{templateType}` that do nothing at all: https://github.com/nette/latte/commit/4c5f3fa69f33ff1a8027d574a7dd27b2cb566ef7

mabar:
Good enough for plugin support 👍

dg:
Annotations `@var` generating will be more difficult because it must be done in individual methods.

mabar:
Yeah, I know. First class support for static analysis will be way more difficult, it was more like a concept

PR #213, opened 2020-02-11
Filters: Update `json_encode()` to PHP 7+ style

jakubboucek:
- bug fix / new feature? modernize code
- BC break? no
- doc PR: nette/docs? -  

Package is now require PHP 7, which know strict mode `json_encode()`. This PR swich old error hunting to using strict `json_encode()`. 

To keep BC is native `\JsonException` translated to `\RuntimeException` with same value of `$message` and `$code`.

## Tests result
- All tests passed without error.
- PhpStan found 17 errors, but same errors on `master` and this PR – this PR didn't break anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ after_failure:
 jobs:
     include:
         -   name: Nette Code Checker
+        	php: 7.4
             install:
                 - travis_retry composer create-project nette/code-checker temp/code-checker ^3 --no-progress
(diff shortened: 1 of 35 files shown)


dg:
JSON_THROW_ON_ERROR exists since PHP 7.3 and Latte works with 7.1.

jakubboucek:
Ouch! Thanks, my mistake.

PR #214, opened 2020-02-15
Filters: fix escaping of IHtmlString from nette/utils

xificurk:
- bug fix
- BC break? no
- doc PR: not needed

435139bf5b75f811ffef3e79cc84d4170a7946d0 broke compatibility with nette/utils - this package still uses `I` prefix on interfaces, namely there is `HtmlString` interface, it's still called `IHtmlString`.


xificurk:
The PR is against master, but the bug is present in v2.7 branch as well and the fix should be backported.

dg:
thanks!

PR #220, opened 2020-03-20
<script type="module"> is also javascript context

iiic:
<script type="module">//…</script> would by also be Javascript context ( Compiler::CONTEXT_HTML_JS )

attribute type="module" on script elementy is necessary for using dynamic javascript import ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import )


dg:
Thanks

PR #230, opened 2020-09-06
CachingIterator: do not ask when not in valid state

hrach:
- bug fix: yes
- BC break: no, [but...](https://xkcd.com/1172/)
- doc PR: no

The inner iterator is one iteration ahead and therefore when CachingIterator asks for the next value (key), it uses a current() method on the inner iterator. But the iteration may have finished and calling iterator's current() and key() in invalid state may produce an exception or invalid data.

PHP'S internal ArrayIterator returns a null, but generally speaking the iterators should rather throw. When doing types properly, we [can't even type such interface](https://phpstan.org/r/9857a2bd-bd79-4152-b1a8-ee336733ed15).

This PR asks first if the inner iterator is in a valid state.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
 		"nette/php-generator": "to use tag {templatePrint}"
 	},
 	"conflict": {
-		"nette/application": "<2.4.1"
+		"nette/application": "<3.0"
 	},
 	"autoload": {
 		"classmap": ["src/"]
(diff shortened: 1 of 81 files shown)


hrach:
Test failures seems unrelated.

hrach:
@dg what do you think?

dg:
Seems legit.

hrach:
Thank you :)

PR #249, opened 2021-01-05
Snippet name can start or end with quotation mark.

janbarasek:
- bug fix
- BC break? yes

I think the snippet name can start or end with a quotation mark:

Snímek obrazovky 2021-01-05 v 17 39 58

Thanks.


dg:
Fixed in Latte 3

PR #266, opened 2021-04-20
Fix filetime issue on PHP 8

### LINKED ISSUE #265: Throwing an exception in acquireLock(...) fails
Version: 2.10
PHP: 8.0.3

### Bug Description
Throwing an exception at **Engine.php:247** fails as Latte expects the function **error_get_last()** to always return an array from what it tries to read the **message** key. However, the aforementioned function can also return **NULL**, see https://www.php.net/manual/en/function.error-get-last.php
```php
throw new RuntimeException("Unable to create directory '$dir'. " . error_get_last()['message']);
```

### Steps To Reproduce
```php
set_error_handler('myErrorHandler');// <- this is what makes the error_get_last() func to always return NULL instead of an array
$latte = new Latte\Engine();
$latte->setTempDirectory('temp');// <- make sure your app can't access or make that dir!!!
$output = $latte->renderToString('random_template.latte');
```

dakujem:
Bug Fix.

Fixes #265

### DIFF
--- a/src/Latte/Engine.php
+++ b/src/Latte/Engine.php
@@ -10,10 +10,11 @@ declare(strict_types=1);
 namespace Latte;
 
 
+use ErrorException;
 use Latte\Compiler\Compiler;
 use Latte\Compiler\Macro;
 use Latte\Compiler\Parser;
-use Latte\Compiler\PhpHelpers;
+use Throwable;
 
 /**
  * Templating engine Latte.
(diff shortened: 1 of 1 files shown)


dg:
Fix your error handler.

dakujem:
Thanks for this helpful tip.

For anyone else who might stumble upon this, make sure you error handler (passed to `set_error_handler` function) does NOT use a comparison like the following to detect error suppression, since it does not work in PHP 8+, [more processing](https://github.com/getsentry/sentry-php/blob/3.2.1/src/ErrorHandler.php#L275) is needed to make it work
```php
// WRONG in PHP 8 error handler
0 === error_reporting()
```

PR #270, opened 2021-06-11
Engine: Remove unused variables (typo).

janbarasek:
- bug fix
- BC break? no

Typo.


src/Latte/Engine.php
```diff
@@ -217,7 +217,7 @@ private function loadTemplate(string $name): void
 		if ($lock) {
 			flock($lock, LOCK_UN); // release shared lock so we can get exclusive
 		}
-		$lock = $this->acquireLock("$file.lock", LOCK_EX);
+		$this->acquireLock("$file.lock", LOCK_EX);
```
dg, line 220: This is intentional

src/Latte/Engine.php
```diff
@@ -301,7 +301,7 @@ public function addFilter(?string $name, callable $callback)
 	public function addFilterLoader(callable $callback)
 	{
 		$this->filters->add(null, function ($name) use ($callback) {
-			if ($filter = $callback($name)) {
+			if ($callback($name)) {
```
dg, line 304: add($name, $filter) ?

dg:
done

Issue #262, opened 2021-02-18
Propagate type information in compiled template code to allow static analysis


MartinMystikJonas:
I just finished first draft of script that uses PHPStan to satic analysis of compiled Latte templates in our projects.

One think I noticed is that {varType} or {templateType} macros does not propagate any type related information to compiled Latte templates so PHPStan cannot check it.

Is there any plan to address this?

I could try to prepare PR which will address this but I would like to disccuss what approach to take here.

One option would be to simply add relevant `/** @var type $variable` annotations after each extract of parameters. his is only way for 

Another IMHO better approach would be to generate known params as template properties and instead of addressign them as `$variable` use `$this->variable` if variable is known but I am not sure how difficult that change would be.

lulco:
Hi @MartinMystikJonas, @dg

I am thinking about this too...

What about this:

```
/** @var int $a */
$a = $this->params['a'];

/** @var string $b */
$b = $this->params['b'];

/** @var Selection $c */
$c = $this->params['c'];

extract($this->params);

```
where $a, $b, $c annotations are from varType or properties from templateType and could be used for static analysis, and extract($this->params) just keep it working like before this change.

jakubvojacek:
Subscribing, that'd be cool!

jakubvojacek:
Anyone managed to come up with somewhat usable solution how to analyse lattes via phpstan? The `@var` annotation would certainly be helpful

lulco:
@jakubvojacek I don't think so. `@var` annotations are just first part of it. I tried to add them to compiled templates and run phpstan on them but there are many other issues like filters etc.

jakubvojacek:
too bad, that'd be killer feature :(

MartinMystikJonas:
I have managed to solve it. It is just proof of concept at the moment and without var annotations it is missing important checks. But it works. At the momeny I simply use PHpStan stubs to define known filters but I have idea how that can by solved better.

MartinMystikJonas:
If you are intetested I will share my implementation with you guys.

jakubvojacek:
Please, that'd be perfect, I am sure many would appreciate that

lulco:
@MartinMystikJonas I am interested too :)

MartinMystikJonas:
Here you go: https://github.com/MetisFW/phpstan-latte

jakubvojacek:
[...]

PR #275, opened 2021-09-27
Propagate variable types to generated code to allow statical analysis

MartinMystikJonas:
- new feature
- BC break? no

With this change Latte will propagate known types infomation from ```{varType}```, ```{define}``` and ```{parameters}``` into compiled PHP code in form of ```/** @var type $var */``` annotations. This will allow statical anylysis of resulting code by static analysers like PHPStan, Psalm, etc.

See discussion about this feature here https://github.com/nette/latte/issues/262


MartinMystikJonas:
@dg Can you please chek this if it is valid approach from your Point of view? It it seems ok to you I will add some more tests for edge cases and support for extracting types from {templateType}

MartinMystikJonas:
@dg Should I consider that like as approval of this approach? If so can you please approve to run workflows? Thanks.

src/Latte/Macros/BlockMacros.php
```diff
@@ -547,7 +552,7 @@ private function addBlock(MacroNode $node, string $layer = null): Block
 	private function extractMethod(MacroNode $node, Block $block, string $params = null): void
 	{
 		if (preg_match('#\$|n:#', $node->content)) {
-			$node->content = '<?php extract(' . ($node->name === 'block' && $node->closest(['embed']) ? 'end($this->varStack)' : '$this->params') . ');'
+			$node->content = '<?php ' . ($node->name === 'block' && $node->closest(['embed']) ? 'extract(end($this->varStack));' : $this->getCompiler()->paramsExtraction)
```
MartinMystikJonas, line 555: This change is necessary to keep params extraction inside block same as in top-level. Previous version always extracted all params inside block even when {parameters} is used.

src/Latte/Compiler/Compiler.php
```diff
@@ -174,7 +177,8 @@ private function buildClassBody(array $tokens): string
 		$this->output = &$output;
 		$this->inHead = true;
 		$this->macroNode = new RootNode;
-		$this->htmlNode = $this->context = $this->paramsExtraction = null;
+		$this->htmlNode = $this->context = null;
+		$this->paramsExtraction = $this->defaultParamsExtraction;
```
MartinMystikJonas, line 181: Default params extraction needs to be set in the beggining so macros can extend it with type information (prepend @var annotations)

dg:
Hi, please give me some time, I'm busy this week. 

Otherwise, it's best to do PR against the last released version, because I change and forcepush the master a lot.

MartinMystikJonas:
@dg Understood. Take your time and look at it whenever you will have time to check it. Let me know if you have any questions or requests how to finish this PR. 

I will rebase it onto last release.

MartinMystikJonas:
New PR targeted onto v2.10 here #276

PR #276, opened 2021-10-03
Propagate variable types to generated code to allow statical analysis

MartinMystikJonas:
- new feature
- BC break? no

With this change Latte will propagate known types infomation from ```{varType}```, ```{define}``` and ```{parameters}``` into compiled PHP code in form of ```/** @var type $var */``` annotations. This will allow statical anylysis of resulting code by static analysers like PHPStan, Psalm, etc.

See discussion about this feature here https://github.com/nette/latte/issues/262

(Previous PR targeted onto master here #275)

### DIFF
--- a/src/Latte/Compiler/Compiler.php
+++ b/src/Latte/Compiler/Compiler.php
@@ -50,8 +50,11 @@ class Compiler
 	/** @var string[] @internal */
 	public $placeholders = [];
 
-	/** @var string|null */
+	/** @var string */
 	public $paramsExtraction;
+
+	/** @var string */
+	private $defaultParamsExtraction = 'extract($this->params);';
 
 	/** @var Token[] */
 	private $tokens;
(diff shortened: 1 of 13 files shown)


MartinMystikJonas:
I just encountered another bug that would be prevented by this. Any chance you would have time to review this PR?

I know it is a bit lrge changeset so if you are busy a the moment another way that would allow me to use this feature would be to create PR with just changes to Compiler that would allow use of custom macros for this. Would it be better?

zobo:
This feature would be very nice to have. @df Would it help if I port it to the latest branch or master?

dg:
@zobo I plan to add this feature in Latte soon.

MartinMystikJonas:
@dg Let us know if we can help with that

dg:
I put the first draft here #297

MartinMystikJonas:
@dg Would you want me to add new tests based on tests in this PR? IT seems you test only very basic things but my tests covers few edge cases - like forgetting global variable types inside blocks.

MartinMystikJonas:
@dg What woukd be best way to prepare it? PR against analysis branch of your (dg) fork? I noticed you did some force pushes there.

dg:
I don't know how to allow you to push into that branch, but how about you take those two commits here and continue?

MartinMystikJonas:
> I don't know how to allow you to push into that branch, but how about you take those two commits here and continue?

Well maybe best way would be just send you patch with changes. My tests would probably require some tweaking to your implementation. I will look at it and then we will figure out how to put it into your branch.

MartinMystikJonas:
Replaced by #309

PR #282, opened 2022-02-17
feat: allow use of text/plain for script tag

lubomirblazekcz:
- bug fix / new feature?   
- BC break? yes
- doc PR: nette/docs#???  

For GDPR usage most scripts must be set as `text/plain`, but latte doesn't handle types in variables with this type.

eg. here - https://orestbida.com/demo-projects/cookieconsent/

This PR fixes it, but might cause bc break, if anyone used the `text/plain` and inserted quotes manualy

eg.
`Do not place {$test} inside quotes in .../templates/@layout.latte:20`

But this change is definitely needed, because if you would want to make the cookieconsent for scripts conditional, there would be problems with the js variables being or not being processed with latte.


dg:
Thanks

PR #297, opened 2022-05-18
Statical analysis for Latte [WIP]

dg:
The purpose of this PR is to enable statical analysis of resulting code by analysers like PHPStan, Psalm, etc.

Related #262, #275, #276

First commit propagates known types infomation from `{var type}`, `{varType}`, `{define}`, `{templateType}` and `{parameters}` into compiled PHP code in form of `/** @var type $var */` annotations. 

The second commit propagates filter parameter in the form of `@property` annotations and stub class:

```php
/**
* @property FiltersTemplate4d1f38ae82 $filters
*/
final class Template4d1f38ae82 extends Latte\Runtime\Template
{

	public function main(array $ʟ_args): void
	{
		echo LR\Filters::escapeHtmlText(($this->filters->lower)(($this->filters->upper)($a))) /* line 1 */;
	}
}

/**
* @property callable(mixed): string $lower
* @property callable(mixed): string $upper
*/
class FiltersTemplate4d1f38ae82
{
}
```

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,4 +1,8 @@
 parameters:
+	excludePaths:
+		analyse:
+			- src/Latte/Compiler/TagParser.php
+
 	level: 5
 
 	paths:
--- a/src/Latte/Compiler/Block.php
+++ b/src/Latte/Compiler/Block.php
@@ -26,6 +26,7 @@ final class Block
 
 	/** @var ParameterNode[] */
 	public array $parameters = [];
+	public VariableScope $variables;
 
 
 	public function __construct(
--- a/src/Latte/Compiler/PrintContext.php
+++ b/src/Latte/Compiler/PrintContext.php
@@ -77,10 +77,14 @@ final class PrintContext
 	/** @var Escaper[] */
 	private array $escaperStack = [];
 
+	/** @var VariableScope[] */
+	private array $scopeStack = [];
+
 
 	public function __construct(string $contentType = ContentType::Html)
 	{
 		$this->escaperStack[] = new Escaper($contentType);
+		$this->scopeStack[] = new VariableScope;
 	}
 
 
@@ -160,9 +164,28 @@ final class PrintContext
 	}
 
 
+	public function beginVariableScope(): VariableScope
+	{
+		return $this->scopeStack[] = clone end($this->scopeStack);
+	}
+
+
+	public function restoreVariableScope(): void
+	{
+		array_pop($this->scopeStack);
+	}
+
+
+	public function getVariableScope(): VariableScope
+	{
+		return end($this->scopeStack);
+	}
+
+
 	public function addBlock(Block $block): void
 	{
 		$block->escaping = $this->getEscaper()->export();
+		$block->variables = clone $this->getVariableScope();
 		$block->method = 'block' . ucfirst(trim(preg_replace('#\W+#', '_', $block->name->print($this)), '_'));
 		$lower = strtolower($block->method);
 		$used = $this->blocks + ['block' => 1];
--- a/src/Latte/Compiler/TemplateGenerator.php
+++ b/src/Latte/Compiler/TemplateGenerator.php
@@ -40,17 +40,20 @@ final class TemplateGenerator
 		bool $strictMode = false,
 	): string {
 		$context = new PrintContext($node->contentType);
-		$code = $node->main->print($context);
-		$code = self::buildParams($code, [], '$ʟ_args', $context);
-		$this->addMethod('main', $code, 'array $ʟ_args');
+		$scope = $context->getVariableScope();
+		$this->addMethod('main', '');
 
 		$head = (new NodeTraverser)->traverse($node->head, fn(Node $node) => $node instanceof Nodes\TextNode ? new Nodes\NopNode : $node);
 		$code = $head->print($context);
 		if ($code || $context->paramsExtraction) {
 			$code .= 'return get_defined_vars();';
-			$code = self::buildParams($code, $context->paramsExtraction, '$this->params', $context);
+			$code = self::buildParams($code, $context->paramsExtraction, '$this->params', $context, $scope);
 			$this->addMethod('prepare', $code, '', 'array');
 		}
+
+		$code = $node->main->print($context);
+		$code = self::buildParams($code, [], '$ʟ_args', $context, $context->getVariableScope());
+		$this->addMethod('main', $code, 'array $ʟ_args');
 
 		if ($node->contentType !== ContentType::Html) {
 			$this->addConstant('ContentType', $node->contentType);
@@ -100,7 +103,7 @@ final class TemplateGenerator
 					: [$block->method, $block->escaping];
 			}
 
-			$body = $this->buildParams($block->content, $block->parameters, '$ʟ_args', $context);
+			$body = self::buildParams($block->content, $block->parameters, '$ʟ_args', $context, $block->variables);
 			if (!$block->isDynamic() && str_contains($body, '$')) {
 				$embedded = $block->tag->name === 'block' && is_int($block->layer) && $block->layer;
 				$body = 'extract(' . ($embedded ? 'end($this->varStack)' : '$this->params') . ');' . $body;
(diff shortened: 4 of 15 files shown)


JanTvrdik:
Very much looking forward to this. I did some initial testing and the main issue I encountered is, that it does not support "closure" variables, i.e. `$foo` would be reported as undefined in the template bellow.

~~~latte
{define outer, Foo $foo}
	{embed wrapper}
		{block inner}{$foo->text}{/block}
	{/embed}
{/define}
~~~

MartinMystikJonas:
I tried fist attempts to migrate my tests to your implementation. SO far I notived few things I would need to solve before I continue:

1) There are many failing tests currently some of them are tests I want to extend by additional checks. It would be best if you could fix these tests befaore I start adding new checks.
2) When you print variable type comments for multiple variable it is printen on single line like this `/** @var string $a *//** @var string $c */` I think it would be best to print each var on separate line for better readability and also it would simplify some tests expects.
3) When I prepared my implementation I stumbled upon one problem. Tga {varType} does not diffrentiate between specifying global varibles (passed to template) and local variables (created inside template). This becomes problem when you starts to use blocks. Because block see only global variables and not local variables (unless explicitly passed). Your implementation adds all types from {varType} to global scope so it is printet also inside block where these variables are not available. I solved this issue by adding variables to global scope only if {varType} is used in template head and othervise it will just output type comment in place. This is snippet of my code:
```php
$comment = "/** @var $type $variable */\n";
		if ($this->getCompiler()->isInHead()) {
			$this->getCompiler()->paramsExtraction .= $comment;
			return "";
		} else {
			return $writer->write($comment);
		}
```

MartinMystikJonas:
@dg Example of global/local sope test I wrote in my implementation:

```php
$template = <<<'XX'

{varType string $a}

{$a}

{varType string $c}
{var $c = 10}

{include test}

{define test}
  {varType int $b}
  {var $b = 5}
  {$a}{$b}
{/define}

XX;

Assert::matchFile(
	__DIR__ . '/expected/varType.phtml',
	$latte->compile($template)
);
```

Expects vartype.phtml:
```
<?php
%A%
	public function main(): array
	{
		extract($this->params);
		/** @var string $a */
%A%
	public function blockTest(array $ʟ_args): void
	{
		extract($this->params);
		/** @var string $a */
%A%
		/** @var int $b */
		$b = 5%a%;
%A%
```

Notice that in blockTest there is no variable type comment for $c, which is local variable.

MartinMystikJonas:
Main issue currently is with global/local variables as mentioned few comments back. In new latte I did not find any way how to differentiate between them (detection of head) - maybe it should be part of PrintContext?

Also I noticed that {var} and {default} tegs leads to duplicate print of variable comments like this (see result of var.default.phpt):

```
extract($this->params); /** @var A\B $var1 */ /** @var ?A\B $var2 */ /** @var A\B $var1 */$var1 = 123; /** @var ?A\B $var2 */$var2 = 456
```

MartinMystikJonas:
@@dg: Let me know if you have idea how to solve this issues. After we solve it I will add more tests originally created for my branch and we can finish this.

dg:
ad passing variables:

I see a fundamental problem here. We can add annotations `/** @var */ exactly according to the rules of how variables are passed to blocks, and static analysis can check the code. But the second part is completely missing, and that is checking that I am passing the correct varibles when I include a block or template. And actually, I can pass almost anything to them.

In order for templates to be really robustly checked, they should explicitly declare all input variables. This can currently be done for a template using `{parameters}` and for a block using `{define ..., params}`. At the same time, `{varType}` should almost not need to be declared, similar to how `/** @var */` is rarely used in PHP code.

In fact, I'm thinking we need to figure out exactly how this is supposed to work first…

mabar:
Imho `{parameters}` and `{varType}` (at root level)  should be considered just a workaround as they are disconnected from variables passed from PHP. For *strict* settings of analysis I would like to forbid their use and instead require `{templateType}`

`{templateType}` could also enforce usage of its class

`{varType}` make sense only above variable. Imho it should behave same way as `@var` in php, so it allows even types not supported by php itself.

`{var type}` is declaration of inline typed variable. Currently not supported by PHP, but it would probably be `Foo $foo = new Foo();`. To be compatible `{var}` should support only types supported by PHP in position of object property (types supported by parameters and return types differ).
I think it could be implemented via anonymous class loaded from custom php stream, so it may be generated in runtime

dg:
@MartinMystikJonas My point is that PHPStan should be able to check this:

```latte
{var $a = 1}
{block foo}
    {=$a + $a} OK
{/block}

{include block foo, a: new stdClass}
```

JanTvrdik:
Would it be possible to introduce some "Latte fully typed mode" which would change compilation of

```latte
{block description, string $description}...{/block}
```

from

~~~php
public function blockDescription(array $ʟ_args): void
{
    extract(this->params);
    $description = $ʟ_args[0] ?? $ʟ_args['description'] ?? null;
    unset($ʟ_args);
    ...
}
~~~

into

~~~php
public function blockDescription(string $description): void // string $description, ...$others
{
    extract(this->params, EXTR_SKIP);
    ...
}
~~~

and using calling it with `(...$params)` instead of `($params)`?

MartinMystikJonas:
@dg I do not think this would be possible without specifying types in `{define}`. But with types specified in define we would need way to specify them as required parameters. For PHPStan it would be possible using array shapes `$ʟ_args array{a: int}` but it would work only for PHPStan and I do not think we should target PHPStan only.

If we can @JanTvrdik proposal with named parameters work it would be better. 

Honestly I think that current params and args passing style with extract is quite magical and if we want proper static analysis it should be reworked but it will introduce BC breaks. I personally would preffer these changes:

1) templateType should make compiled template extend given template type.
2) template params should be accessed as template properties `$this->paramName`
3) blocks should require definition of all args that can be passed - and throw error when we try to pass arg that is nod specified in block definition
etc.

But this might be better to leave to discussion for some future major version.

MartinMystikJonas:
@JanTvrdik Is there reason why do not make this a default mode? Some possible BC break?

dg:
There is no such dilemma. 

The whole thing is quite complex and needs a deep analysis. It is really pointless to deal with implementation details, they are not important now. What is important is how the variable passing is supposed to work from the point of view of the template designer.  It is necessary to preserve backwards compatibility, but it is possible to switch between the two modes, for example by the presence of the `{parameters}` tag for example.

mabar:
Reserving some variable like `$params` and contain variables from `Template` class might be easier, `$this` is already used. Also proxy Template classes would need Template implementation to be not final to be extendable.

@dg If you would agree on my idea that `{varType string $f} == /** @var string $f **/` and `{var string $f} == string $f; ` then in same way could exist `{templateType FooTemplate}` and `{template FooTemplate}`. One with just hints, second activating strict mode.

With `{template FooTemplate}` would be activated strict mode. Instead of extracting variables from it, reserved variable `$params` would contain whole `FooTemplate`

MartinMystikJonas:
I would preffer going for minimal viable solution now:
1) Support only PHP native types
2) All outputs only in form of comments (to prevent accidental BC breaks caused by type checks)
3) Accept that come checks (like @dg example of calling block) would not be possible yet

And then we can think about some rework of params/args handling (maybe with some possible BC breaks) in the future that would allow better checks.

JanTvrdik:
> My point is that PHPStan should be able to check this

I don't see an easy way to do this as long as templates are compiled independently of each other.

MartinMystikJonas:
Also when rendering of block uses indirect magic not direct method call:
```
$this->renderBlock('test', get_defined_vars(), 'html')
```

To be able to check this by any static analysis tool we would first need major change of how templates are compiled to get rid of this indirect magic and transform it to direct method call with params like:
```
$this->blockTest('html', $someParam)
```
Without it I do not think it is possible to do this kind of static analysis.

MartinMystikJonas:
It seems we somehow frozen here :-)

@dg What can we do to move this forward? I would really like to have at least some basic version available for work I have planned for this summer.

dg:
It turns out that it can either be done quickly and relatively easily, but very imperfectly, or perfectly :-)

Unfortunately, I am a perfectionist and I choose the more difficult path.

MartinMystikJonas:
Well can't we have both? First imperfect way that covers majority of errors and later perfect way that enables covering more subtle errors? ;-)

dg:
I'd be glad if you'd create the "imperfect" one.

MartinMystikJonas:
Would be happy to do it. 

How to do it? It would be best to start from your analysis branch and create new PR to master based on it?

dg:
I think that's best.

PR #309, opened 2022-07-11
Statical analysis for Latte [WIP] by MartinMystikJonas

MartinMystikJonas:
Continue of work started by @dg in https://github.com/nette/latte/pull/297


MartinMystikJonas:
Ok I think I have functional version that passes all tests I used in my previous PR.

I also changed behaviour of `{parameters}` tag so it also applies inside blocks. Instead of extracting `$this->params` it now extracts `$this->prepare()` and adds definition of types used in `paramsExtraction` (see `tags/excpeted/parameters.inc8.phtml`)

MartinMystikJonas:
@dg One think I am not sure of is double meaning of `varType`. When used in template head it is interpreted as definition of type of parameter (so this type information is propagated inside blocks) and when used outside of head it is interpreted as definition of type of local variable.

I am not sure if this double meaning is ok. Maybe we should make `{varType}` just about local variables types and add something like `{parameterType}` to define types of template paramater. What do you think?

MartinMystikJonas:
@dg @lulco @jakubvojacek @JanTvrdik @mabar Guys, could you check this version and let me know what do you think? I would like to finish this basic version.

lulco:
Like I wrote somewhere, I'm using https://github.com/revealphp/reveal-src (previously in symplify) for some months maybe year already and it works like a charm for my projects.

MartinMystikJonas:
@dg Hi. I would like to finish this feature. Is there anything I can do (aside from resolving that new conflict) to have this merged? I do not want to hurry you I just want to know if there is anything I can do.

MartinMystikJonas:
@janzimmermannnetrex I would like to see that rule.

janzimmermannnetrex:
@MartinMystikJonas see https://drive.google.com/file/d/1YiJIllqQH3Iyb0rns-rYl2vLX_ro05IN/view?usp=sharing for inspiration. It is definitely not ready to use it as is in another project yet. 
* We are trying to create a presenter instance during the static analyses (you need to include nette bootstrap in phpstan)
* We try to call the `link()` on presenter instance. It throws an exception once the link is not valid.
* It is similar for Latte as well.
Sure I'd appreciate any ideas to improve it.

MartinMystikJonas:
@survilla I implemented similar functionality for phpstan-nette with a bit more features and as static-analysis (without actually calling link()). You can check it here if you are interested: https://github.com/phpstan/phpstan-nette/pull/132

PR #312, opened 2022-08-07
callables of Extension::getTags() can return null

MartkCz:


dg:
What is the point of distinguishing between null and void?

Issue #314, opened 2022-08-22
[Feature Request] Support for multiple paths


alexander-schranz:
Currently Latte does only provide a `FileLoader` which allows to bind templates to a single directory.

I think it would be great if Latte would provide additional Loader which allows to register multiple directories via a namespace. So Latte could example use in Laminas Framework as Renderer also by multiple Modules which could registering additional paths.

In twig example the loader accepts multiple paths and the loader will then look one after the other directory.

Another possibility in twig is a namespace example I can have `@app/test.html.twig` and `@other/test.html.twig` and register paths like:

```php
[
    'app' => __DIR__ . '/templates',
    'other' => __DIR__ . '/vendor/other/module/templates',
]
```

While in `twig` the `@` symbol is used, I found while working on [my abstraction](https://github.com/schranz-templating/templating) that other frameworks use the `::` syntax for this e.g. `other::blade`.

A implementation could look like the following:

<details>

<summary>MultiPathLoader.php</summary>

```php
<?php

/**
 * This file is part of the Latte (https://latte.nette.org)
 * Copyright (c) 2008 David Grudl (https://davidgrudl.com)
 */

declare(strict_types=1);

namespace Latte\Loaders;

use Latte;


/**
 * Template loader.
 */
class MultiPathLoader implements Latte\Loader
{
    private array $loaders = [];

	public function __construct(?array $baseDirs = ['' => null])
	{
        foreach ($baseDirs as $key => $baseDir) {
            $this->loaders[$key] = new FileLoader($baseDir);
        }
	}


	/**
	 * Returns template source code.
	 */
	public function getContent(string $name): string
	{
        [$loader, $name] = $this->extractLoaderAndName($name);

        return $loader->isExpired($name, $name);
	}


	public function isExpired(string $file, int $time): bool
	{
        [$loader, $name] = $this->extractLoaderAndName($file);

        return $loader->isExpired($name, $time);
	}


	/**
	 * Returns referred template name.
	 */
	public function getReferredName(string $name, string $referringName): string
	{
        [$loader, $name] = $this->extractLoaderAndName($name);

        return $loader->getReferredName($name, $referringName);
	}


	/**
	 * Returns unique identifier for caching.
	 */
	public function getUniqueId(string $name): string
	{
        [$loader, $name] = $this->extractLoaderAndName($name);

        return $loader->getUniqueId($name);
	}


    private function extractLoaderAndName(string $name): array
    {
        if (\str_starts_with('@', $name)) {
[...]

PR #320, opened 2022-11-01
Add namespace loader

alexander-schranz:
- bug fix / new feature?  #314
- BC break? no
- doc PR: nette/docs#???  

This introduce a namespace loader. This allows to use a namespace like `app::` similar to Plates and Blade template engine.

Usage:

```php
    $defaultLoader = new FileLoader(__DIR__ . '/templates');
    $emailLoader = new FileLoader(__DIR__ . '/emails');
    $someLoader = new FileLoader(__DIR__ . '/vendor/some/some/templates');
    $otherLoader = new StringLoader(['main' => 'othercontent']);

    $loader = new NamespaceLoader([
        '' => $defaultLoader,
        'email' => $emailLoader,
        'some' => $someLoader,
        'other' => $otherLoader,
    ]);
```

This way it is very flexible as we can use any loader again.


alexander-schranz:
@dg looks like you rebased it. Do I need to change the target branch?

Crell:
It looks like @dg has been rebasing this diligently for 2 years.  I am also very interested in this feature (the thing I'm trying to do right now likely won't really work without it), as currently there seems to be no way to reference a layout that is not in the same directory as the template, or else have relative paths defined in the {layout} block (ew).

Is there something we users could do to help move this forward, or get attention on it?

PR #353, opened 2023-11-17
feat: add support for application/ld+json

### LINKED ISSUE #352: Escaping `application/ld+json`?
Version: 3.0.12

### Bug Description

I did not find a way how to force latte to escape double quotes in a json string. There's `noescape` helper but no `escape` helper or sth. Docs says latte should escape automatically but it does not seem to do so.

### Steps To Reproduce

https://fiddle.nette.org/latte/#f147ffe9e0

```latte
{var $text = '"lorem" ipsum'}
<script id="json-ld-script" type="application/ld+json">
    {"text": "{$text}"}
</script>
```

It generates invalid json `{"text": ""lorem" ipsum"}`

### Expected Behavior
Escaped double quotes

### Possible Solution

Escape `"`

simPod:
- bug fix / new feature?   Fixes #352
- BC break? yes
- doc PR: nette/docs#???  

Adds js escaping to application/ld+json

### DIFF
--- a/src/Latte/Compiler/Escaper.php
+++ b/src/Latte/Compiler/Escaper.php
@@ -120,7 +120,7 @@ final class Escaper
 			if ($el->is('script')) {
 				$type = $el->getAttribute('type');
 				if ($type === true || $type === null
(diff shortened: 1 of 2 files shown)


dg:
Can you please change it to `application/.*\+json`

simPod:
done

PR #395, opened 2025-05-16
Support property hooks in template params

### LINKED ISSUE #393: Support property hooks in template params
Version: 3.0.20

### Bug Description

It seems that when I use an object for template parameters, property hooks are not properly supported at this moment.

### Steps To Reproduce

```php
class Params
{
    public function __construct(public string $firstName, public string $lastName) {}
    public string $fullName { get => $this->firstName . ' ' . $this->lastName; }
}

$params = new Params('John', 'Doe');
$html = $latte->renderToString('template.latte', $params);
```

Template:

```latte
{$fullName}
```

This throws `Undefined variable $fullName`.

### Expected Behavior

No error, display the value.

### Possible Solution

I believe [this code](https://github.com/nette/latte/blob/7b397b4fb525613134240248897f39b13e35a88e/src/Latte/Engine.php#L543) is the culprit because it casts the object to an array, which omits the virtual property.

As per the [property hooks RFC](https://wiki.php.net/rfc/property-hooks#interaction_with_serialization), this seems to be by design:

> When casting an object to an array (`$arr = (array) $obj`), currently the visibility of properties is ignored; the keys returned may have an extra prefix in them to indicate that they were private, but that's it. As this operation currently reveals internal implementation details, it also will not invoke the get hook.

This might also have implications for non-virtual properties with a `get` hook, because I would intuitively expect the hook to be used when I reference the variable in the template.

Perhaps the code should use reflection to list public properties and get their values in a way that does not bypass the `get` hook?

jiripudil:
- bug fix, closes #393 
- BC break? no
- doc PR: none

### DIFF
--- a/src/Latte/Engine.php
+++ b/src/Latte/Engine.php
@@ -519,7 +519,8 @@ class Engine
 			return $params;
 		}
 
-		$methods = (new \ReflectionClass($params))->getMethods(\ReflectionMethod::IS_PUBLIC);
+		$reflection = new \ReflectionClass($params);
+		$methods = $reflection->getMethods(\ReflectionMethod::IS_PUBLIC);
 		foreach ($methods as $method) {
 			if ($method->getAttributes(Attributes\TemplateFilter::class)) {
 				$this->addFilter($method->name, [$params, $method->name]);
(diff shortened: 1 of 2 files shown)


dg:
Thanks

PR #415, opened 2026-04-09
Fix PhpHelpers::checkCode crash when PHP binary path contains spaces

davidkvasnovsky:
## Problem

`PhpHelpers::checkCode` passes the PHP binary path to `proc_open` as a concatenated string:

```php
$phpBinary . ' -l -n'
```

On macOS with Laravel Herd, `PHP_BINARY` resolves to a path containing spaces:

```
/Users/.../Library/Application Support/Herd/bin/php84
```

The shell splits this at the space, resulting in exit code 127 (`No such file or directory`), empty stdout, and then a crash at line 271:

```
Warning: Undefined array key 1 in PhpHelpers.php on line 271
Fatal error: strip_tags(): Argument #1 ($string) must be of type string, null given
```

This affects the Latte linter (`Latte\Tools\Linter`) which enables the PHP linter via `enablePhpLinter(PHP_BINARY)`.

## Fix

Pass an array to `proc_open` instead of a string. This bypasses the shell entirely, avoiding word-splitting. This is also what the existing `bypass_shell` option was intended to do, but that only works on Windows — the array form works on all platforms.

```diff
-$phpBinary . ' -l -n',
+[$phpBinary, '-l', '-n'],
```


dg:
thanks

PR #419, opened 2026-06-16
Implicit default block for embeds

### LINKED ISSUE #417: make content of {embed} available
Thank you for this great templating engine, I love it!

It would be great if we could have something like this:

```
{define myblock, $var}
    some markup
    {$var}
    some more markup
    {block content}default content{/block}
    some other markup
{/define}

{embed myblock, "hello"}
    world
{/embed}
```
Note: "world" is not wrapped in a `{block content}` tag.

This should render as:

```
    some markup
    hello
    some more markup
    world
    some other markup
```

Here, `content` is used as a keyword (maybe something like `__content` is better to avoid conflicts).

With the current version of latte I have to wrap "world" in a block which makes templates harder to read.

Regards,
Henrik

daun:
- New feature, closes #417
- BC break? no
- doc PR: happy to write one if you're on board with this feature :)

## What this does

Provide an implicit `default` block for `embed`s and move any "loose" children into that default block.

Mirrors how other templating libraries deal with default/unnamed slots.

```latte
{* Pass button label into implicit `default` block *}

{embed button}
  Submit
{/embed}
```

## Current state

Currently, I manually have to define the default block when embedding.

```latte
{define button}
  <button>
    {block default}Click{/block}
  </button>
{/define}

{embed button}
  {block default}
    Submit
  {/block}
{/embed}
```

It's a lot of noise when using many embeds.

```latte
<ul>
  <li>{embed button} {block default} Submit {/block} {/embed}</li>
  <li>{embed button} {block default} Reset {/block} {/embed}</li>
  <li>{embed button} {block default} Cancel {/block} {/embed}</li>
</ul>
```

## With this feature

This will move any "loose" children into an implicit `default` block that the template file can render as necessary. 

```diff
 {embed button}
-   {block default}
     Submit
-   {/block}
{/embed}
```

```latte
<ul>
  <li>{embed button}Submit{/embed}</li>
  <li>{embed button}Reset{/embed}</li>
  <li>{embed button}Cancel{/embed}</li>
</ul>
```

Explicit blocks and the implicit default block can live side by side.

```latte
{define button}
  <button>
    {block default}Click{/block}
    {block icon}<svg />{/block}
  </button>
{/define}

{embed button}
  Submit
  {block icon}<svg />{/block}
{/embed}
```

Explicit and implicit `default` block can never coexist and end up with a new compile error.

```latte
{* Throws: "Cannot combine loose content with an explicit {block default} inside {embed}"  *}

{embed button}
  Submit
  {block default}Send{/block}
{/embed}
```

## Tests

Tests are added.

## Nonbreaking

No existing tests have been modified.

Templates currently using a `default` block will keep working.

Also, this does not automatically render the loose content. Loose content is still ignored, just as before. It requires defining a `{block default}` inside the embedded content.

## Alternative implementation

I was thinking of using an unnamed block instead of a default block. Not sure if that would've worked even? Also, I think `default` reads more intentional and in line with how most templating engines name their default slots.

```latte
{define button}
  <button>
    {block}Click{/block}
    {block icon}<svg />{/block}
  </button>
{/define}
```

### DIFF
--- a/src/Latte/Essential/Nodes/EmbedNode.php
+++ b/src/Latte/Essential/Nodes/EmbedNode.php
@@ -8,16 +8,19 @@
 namespace Latte\Essential\Nodes;
 
 use Latte\CompileException;
+use Latte\Compiler\Block;
 use Latte\Compiler\Nodes\FragmentNode;
 use Latte\Compiler\Nodes\Php\Expression\ArrayNode;
 use Latte\Compiler\Nodes\Php\ExpressionNode;
+use Latte\Compiler\Nodes\Php\ModifierNode;
 use Latte\Compiler\Nodes\Php\Scalar\StringNode;
 use Latte\Compiler\Nodes\StatementNode;
 use Latte\Compiler\Nodes\TextNode;
 use Latte\Compiler\PrintContext;
 use Latte\Compiler\Tag;
 use Latte\Compiler\TemplateParser;
-use function count, preg_match;
+use Latte\Compiler\Token;
+use function count, preg_match, trim;
 
 
 /**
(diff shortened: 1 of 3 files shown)


dg:
Thanks!

jan-herman:
Great feature! I've been hoping for this for a while! 

I've just noticed a small bug though. Using `{include parent}` results in the error "Cannot include undefined parent block 'content'."

daun:
@jan-herman Interesting. It's probably best if you open a new issue on the repo, along with the exact latte template that triggers the error for you.

dg:
@jan-herman fixed

